PR #37, opened 2017-06-29
Add Cache Busting in Production Code

ryanstrat:
When Laravel Mix is run in production, the compiled CSS and JS files will be hashed to version them.

Docs: https://github.com/JeffreyWay/laravel-mix/blob/master/docs/versioning.md

This might be adding complexity that is not needed, so I want to get thoughts before we merge it in.


kberzinch:
This doesn't seem to complicate things too much, given that we have to run webpack per deploy anyway. What are you concerned about exactly?

ryanstrat:
This is definitely needed. Ran into browser caching issues on staging environment

Issue #14, opened 2017-06-20
Write API endpoint for Interacting with Users model


ryanstrat:
This will require adding the following methods

- [x] GET one user
- [x] GET all users
- [x] POST new user
- [x] PUT update user and replace all existing attributes
- [x] PATCH update user (only included fields) [patch semantics]
- [x] DELETE soft delete a user

ryanstrat:
Closed via #77

PR #77, opened 2017-07-09
Add User API Endpoints

RossIV:
For Issue #14

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -69,7 +69,7 @@ class UserController extends Controller
      */
     public function show($id, Request $request)
     {
-        $user = User::find($id);
+        $user = $this->getUserByIdentifier($id);
         if ($user) {
             return response()->json(['status' => 'success', 'data' => ['user' => $user]]);
         } else {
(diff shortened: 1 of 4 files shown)


kberzinch:
Closes #14

app/Http/Controllers/UserController.php
```diff
@@ -0,0 +1,156 @@
...
+        ]);
+
+        try {
+            $user = User::create($request->all());
+        } catch (QueryException $e) {
+            $errorMessage = $e->errorInfo[2];
+            return response()->json(['status' => 'error', 'message' => $errorMessage], 500);
+        }
+
+        if (is_numeric($user->id)) {
```
ryanstrat: If we lookup a user by number, should that be GTID rather than their internal ID number?

Perhaps we enable that via a query string?

app/Http/Controllers/UserController.php
```diff
@@ -0,0 +1,156 @@
...
+            $dbUser = User::findOrFail($user->id);
```
ryanstrat: If the user already exists, this will throw an error that is not handled.

app/Http/Controllers/UserController.php
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    /**
+     * Display the specified resource.
+     *
+     * @param  int $id
+     * @return \Illuminate\Http\Response
+     */
+    public function show($id, Request $request)
```
ryanstrat: Same as above, I think we want to be looking up users by GTID or UID.

app/Http/Controllers/UserController.php
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @param  int $id
+     * @return \Illuminate\Http\Response
+     */
+    public function update(Request $request, $id)
+    {
+        if ($request->method() == "PUT") {
+            //Replace all existing fields
+            $this->validate($request, [
```
ryanstrat: Based on my understanding of the laravel docs, I don't think that this is implementing PUT vs PATCH semantics properly.  I'll test when I get home and update this review.

app/Http/Controllers/UserController.php
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Remove the specified resource from storage.
+     *
+     * @param  int $id
+     * @return \Illuminate\Http\Response
+     */
+    public function destroy($id)
+    {
+        $deleted = User::destroy($id);
```
ryanstrat: Add options to delete via GTID or UID

app/Http/Controllers/UserController.php
RossIV: Here are two ways how I see this can be resolved, let me know which of these you'd prefer or if you'd like something different.
1) Use a query parameter such as `/api/v1/users/?gtid=900000000`
2) Auto-detect DB ID or GTID by checking (a) value length and (b) first digit [9] so that any nine-digit string that begins with a nine would assume GTID, anything else would be DB ID.

kberzinch: I vote query parameter, guessing type smells bad

app/Http/Controllers/UserController.php
RossIV: Same as above, this can be done with a parameter or value analysis. Let me know which you'd prefer.

app/Http/Controllers/UserController.php
ryanstrat: I would make GTID or UID options, determined based on int or string and make id an optional query param. The db ID is unlikely to every be used off the server, since we're enforcing UID and GTID as unique keys. I would also be comfortable falling back to Id if the number format is not a GTID formatted number.

app/Http/Controllers/UserController.php
RossIV: PUT is a full replacement of the resource, PATCH is only partial editing - at least from what I've read online. Laravel handles the two methods the same in terms of routing when using resource controller routing if that helps clear things up.

ryanstrat: Conversation in Slack lead to the removal of the implementation of PUT. Both PUT and PATCh to this route will implement patch semantics.

app/Http/Controllers/UserController.php
RossIV: After testing, if you try to create a duplicate user, the error validation picks up on it (as designed) and handles the error like any other validation problem by returning JSON with the errors described.

Issue #45, opened 2017-07-01
Implement basic event management


kberzinch:
Users should be able to CRUD events. Events can optionally have RSVP (and invite) and/or attendance tracking functionality.

ryanstrat:
A RESTful API for this now exists, but a management GUI is needed.

ryanstrat:
Still needs an "add event" form. Which is probably as simple as refactoring the edit form into a component.

RossIV:
Create Event added in #285, delete button added in 221ae7f.

RossIV:
What is left to do here?

ryanstrat:
I think this is done.

PR #92, opened 2017-08-06
Added basic General Interest invite design

### LINKED ISSUE #19: Design General Interest Announcement Email

kberzinch:
This pull request is dependent on #45 for RSVP links and closes #19.


kberzinch:
You should also definitely run this on your own machine to see how it renders - there are some non-obvious additions by default, although we can customize those as well if desired.

app/Mail/GeneralInterestInvite.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Build the message.
+     *
+     * @return $this
+     */
+    public function build()
+    {
+        $this->from('noreply@my.robojackets.org', 'RoboJackets')
```
ryanstrat: Can we set this info in a config file instead?

app/Mail/GeneralInterestInvite.php
```diff
@@ -0,0 +1,40 @@
...
+             ->subject('RoboJackets General Interest Meeting - RSVP Requested')
+             ->markdown('mail.generalinterest.invite');
+
+        $this->withSwiftMessage(function ($message) {
+            $message->getHeaders()
+                    ->addTextHeader('Reply-To', 'RoboJackets Support <support@robojackets.org>');
```
ryanstrat: Sending from a no-reply address seems weird to me if the reply-to is set differently. I don't know what the standards for this are exactly, though

app/Mail/GeneralInterestInvite.php
kberzinch: There's a default configuration in .env that is correct for most mail, but I want to send GI invites as "RoboJackets" not "MyRoboJackets".

app/Mail/GeneralInterestInvite.php
kberzinch: We could leave off the Reply-To. As a general rule, nobody should be replying to this email, but if they do I want it to go to the admin mailbox, hence the setup.

Alternatively we can send from support@my, which also goes to the admin mailbox.

ryanstrat:
Have we dealt with the headers / footers that needed to be removed yet?

PR #99, opened 2017-08-27
FASET Edit form

ryanstrat:
Currently enables viewing and editing name and email for a given FASET entry.

Survey data was left out as it is non-critical, and kinda hard.


ryanstrat:
With approval from @kberzinch in slack, I'm going to merge the PR, but still accepting commentary

PR #103, opened 2017-08-29
User Admin interfaces

ryanstrat:
This is a pretty barebones implementation of and admin UI. Currently there is no client side validation.


ryanstrat:
This is ready for code review.

There is a known issue (#105)

PR #113, opened 2017-08-30
RSVP functionality is ready for General Interest!

ryanstrat:


app/Event.php
```diff
@@ -0,0 +1,28 @@
...
+    ];
+
+    public function organizer()
+    {
+        return $this->hasOne('App\User');
+    }
+
+    public function rsvp()
+    {
+        return $this->belongsTo('App\Rsvp');
```
RossIV: (discussed in Slack) should be a hasMany()

app/User.php
```diff
@@ -42,6 +42,16 @@ public function teams()
...
     }
 
+    public function organizes()
+    {
+        return $this->belongsToMany('App\Event');
+    }
+
+    public function rsvps()
+    {
+        return $this->belongsToMany('App\Rsvp');
```
RossIV: I think this should be hasMany as well

ryanstrat: Fixed

app/Event.php
ryanstrat: fixed

Issue #123, opened 2017-08-30
Make a User Look-up component


ryanstrat:
Throughout the app, there are many places where we need a form field that holds the value of a user ID or UID. However, our users want to be able to find users by typing their name.

Users should be able to type a name, gtid, or uid into the box and receive a typeahead-like list of matching Users. The vue component will emit the either uid or the id of the user, based on a config set by a prop. The default should be to emit a uid.

ryanstrat:
https://github.com/matiastucci/vue-input-tag

RossIV:
I have this mostly implemented with http://sagalbot.github.io/vue-select/, just trying to figure out a bug I'm running into with debouncing.

RossIV:
This is mostly complete. It emits the selected user object, rather than just one attribute which IMO is more flexible. Only thing left is to tweak the filter logic to display names when searching with a GTID or username. OOTB it doesn't do that.

PR #142, opened 2017-09-06
Upgrades to Event Management Page

ryanstrat:
The event detail page now successfully shows RSVPs and the data is editable.

Organizer is currently shown just as id from the database, which will remain until #123 is completed.

### DIFF
--- a/app/Http/Controllers/EventController.php
+++ b/app/Http/Controllers/EventController.php
@@ -73,7 +73,8 @@ class EventController extends Controller
      */
     public function show($id, Request $request)
     {
-        $event = Event::with(['rsvps', 'organizer'])->find($id);
+        $event = Event::with(['rsvps'])->find($id);
+
         if ($event) {
             return response()->json(['status' => 'success', 'event' => $event]);
         } else {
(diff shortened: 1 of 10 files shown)


ryanstrat:
Works on #45 (But does not complete)

app/Http/Controllers/EventController.php
```diff
@@ -90,7 +91,37 @@ public function show($id, Request $request)
      */
     public function update(Request $request, $id)
     {
-        return  response()->json(['status' => 'error', 'message' => 'method_not_implemented'], 501);
+        $event = Event::find($id);
+
+        if (isset($request->organizer)) {
+            $organizer = UserController::getUserByIdentifier($request->organizer);
```
RossIV: After merge of #158, please change this to reflect the newly-created Eloquent scope for searching by identifier.

app/Http/Controllers/EventController.php
```diff
@@ -90,7 +91,37 @@ public function show($id, Request $request)
...
+            $request['organizer'] = $organizer->id;
+        }
+
+        
+        $this->validate($request, [
+            'name' => 'required|max:255',
+            'price' => 'numeric',
```
RossIV: Will numeric allow for decimals? I don't know why it wouldn't, but just checking in case prices are not round dollar amounts.

ryanstrat: Yes, this allows decimals.

The `Integer` validator rule is used otherwise.

ryanstrat:
Requested changes made. Re-review please.

Issue #150, opened 2017-09-07
Add "Active Member" flag to user model


kberzinch:
Dues will have effective start and end dates. These should be queried to determine whether a member is considered "Active".

RossIV:
We should, theoretically, be able to accomplish this with an Eloquent local scope, but I'll have to see how to do this in a manner that isn't completely gross. Pending further investigation.

RossIV:
Referencing notes from our last meeting, I want to make sure that the following assumption is correct for a member being 'active':

Member has made a payment for dues for a term (or in Apiary terms a `DuesPackage`). For that `DuesPackage`, the current date is between the `effective_start` and `effective_end`.

RossIV:
@ryanstrat Approved the above statement. Code added in 79b7d50.

PR #156, opened 2017-09-12
Add Dues/Payment Backend

RossIV:


ryanstrat:
I updated the ER-Diagram for what we think this should look like. Snapshot below

image

app/Http/Controllers/DuesPackageController.php
```diff
@@ -0,0 +1,120 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function update(Request $request, $id)
+    {
+        $this->validate($request, [
+            'id' => 'required|exists:dues_packages',
+            'name' => 'string',
+            'eligible_for_shirt' => 'boolean',
+            'eligible_for_polo' => 'boolean',
+            'effective_start' => 'date',
```
ryanstrat: Does the `date` filter imply `not null`?

app/Http/Controllers/DuesPackageController.php
```diff
@@ -0,0 +1,120 @@
...
+            'effective_end' => 'date',
+            'cost' => 'numeric'
+        ]);
+
+        $package = DuesPackage::find($id);
+        if ($package) {
+            $package->update($request->all());
+        } else {
+            return response()->json(['status' => 'error', 'message' => 'DuesPackage not found.'], 404);
```
ryanstrat: I want to push towards making the status methods consistent error codes in the style of the slack API. I've done it for the APIs I wrote, and we should go back and do the others.

This one could be `dues_package_not_found` or simply `entity_not_found`

ryanstrat:
@RossIV I'm working through implementing some front-end components related to this PR. It would be nice if I had an easy-to-hit frontend that only returned "active" `dues_packages`

For the admin UI, we probably want to show all of them, but for anything user-facing, we only want enabled/currently relevant packages.

RossIV:
@ryanstrat Added column as requested in 65ba1f5. You can use `DuesPackage::active()->get();` to retrieve all active packages.

app/Http/Controllers/DuesPackageController.php
RossIV: To the best of my knowledge, any validation rule will yell at you if you try to pass null without a `nullable` rule in the list. If you add `nullable`  to the validation, you can pass any field with a null value and the model will not be changed. I can add `nullable` to these if you want. That would be consistent with most of the other API validation.

app/Http/Controllers/DuesPackageController.php
RossIV: Let's discuss this further - not sure I see the benefit but if you want it I'm happy to change it.

RossIV:
Wait...I just re-thought your question about 'active' packages @ryanstrat. Did you mean 'active' as in the current date is between the effective dates OR a manual 'active'/'not active' setting, or a combination thereof?

ryanstrat:
I mean 'active' in the sense that is will be displayed to members as a dues purchase option. Not that the membership itself is active

RossIV:
Alright so I wasn't completely off base. I'd propose changing the attribute to 'available_for_purchase' to be more clear on its specific purpose. 'active' could be misconstrued to mean that it's a package that is for the current date period and thus a member with that package would be considered active. Perhaps a derived virtual attribute is in order for that.

RossIV:
Here's my set of suggestions:
1) Change the `active` column to `available_for_purchase` - a boolean that's manually set by administrators to toggle customer-facing visibility of the model. This goes with (2) below.

2) Create a local scope for the `DuesPackage` model - `available_for_purchase` - to restrict DB queries with `->available_for_purchase()` modifier to return models that have the respective attribute set to `true`.

3) Create a local scope for the `DuesPackage` model - `active` - to restrict DB queries with `->active()` modifier to return models where the current date is between the effective start/end dates.

4) Create a virtual attribute for the `DuesPackage` model - `is_active` - that is a boolean for whether or not the current date is between the model's effective start/end dates. This will be returned any time the model is requested via a DB query.

(1) and (2) should solve @ryanstrat 's original request. (3) and (4) will be useful for #150 in determining if a user has paid dues for a package that's active for the current date.

I'm open to suggestions for any changes to these items - naming or otherwise. Just LMK.

ryanstrat:
@RossIV Those all sounds like good changes. What API changes are needed to trigger the new scopes? Perhaps adding `/active` to the end of the URI?

RossIV:
@ryanstrat Since Fractal isn't implemented yet, we could either do a suffix (`/active`) or a parameter (`users/?active=true`). I'm game for either.

ryanstrat:
I think the suffix is the cleaner solution. In other parts of the program, we add a suffix to represent the user or something that selects a specific entity. It's not unreasonable to expect that it could also describe a subset.

RossIV:
All requested changes have been made. Please re-review at your convenience @ryanstrat .

Notes:
- I added the active user attribute into this branch for convenience, and it's related so there's that.
- At the moment there isn't a way to get every active user in a single query - I'm working on that. You can tell if a given user is active, though, by the use of the `is_active` virtual attribute that's returned upon retrieval of the model.
- Active DuesPackages can be retrieved at `api/v1/dues/packages/active`

As always, let me know if you have any questions/concerns/change requests.

RossIV:
Closes #150 
Closes #160

Issue #152, opened 2017-09-07
API JWT authentication is not working


kberzinch:
Likely an issue with chaining middlewares.

RossIV:
Closed in #169

PR #169, opened 2017-09-28
Fix JWT #152

RossIV:
Fixes/Resolves/Closes/Magic Word's #152 

Change JWT to use `uid` instead of `id` which magically fixed things

### DIFF
--- a/config/jwt.php
+++ b/config/jwt.php
@@ -82,10 +82,10 @@ return [
     |
     | Specify a unique property of the user that will be added as the 'sub'
     | claim of the token payload.
-    |
+    |`
     */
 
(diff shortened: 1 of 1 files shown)


kberzinch:
Fixes #152

config/jwt.php
```diff
@@ -82,10 +82,10 @@
     |
     | Specify a unique property of the user that will be added as the 'sub'
     | claim of the token payload.
-    |
+    |`
```
kberzinch: ?

RossIV: uhh..huh.

RossIV:
fixed @kberzinch

PR #171, opened 2017-10-02
Add Authorization, Polymorphic Payments

RossIV:
Adds authorization to all API routes with scoped permissions and roles
Adds polymorphism to the Payment model


app/Http/Controllers/FasetVisitController.php
```diff
@@ -81,6 +97,14 @@ public function update(Request $request, $id)
             return response()->json(['status' => 'error', 'message' => 'visit_not_found'], 404);
         }
 
+        $requestingUser = $request->user();
+        $requestedUser = $visit->user;
+        //Enforce users only updating themselves (update-users-own)
+        if ($requestingUser->cant('update-faset-visits') && $requestingUser->id != $requestedUser->id) {
+            return response()->json(['status' => 'error',
+                'message' => 'Forbidden - You do not have permission to update this FasetVisit.'], 403);
```
ryanstrat: I mentioned this in a prior pr, but I think we shouldn't make our error messages "sentence" style. We should prefer terse text-based error codes.

That error is far to detailed to ever dump to the UI, so I think the target audience is a program, rather than a person.

PR #172, opened 2017-10-02
Member-Facing Dues Workflow

### LINKED ISSUE #149: Add User-Facing Dues Workflow
After #28 (Build Landing Page) is done,

It will lead to the Dues Workflow, which contains the following pages
- [x] Dues Info (Critical info)
- [x] Safety Agreement
- [x] Non-Critical Info
- [x] Demographics collection
- [x] Payment Instructions

ryanstrat:
Closes #149 

Contains a 5-step Dues Payments workflow to collect Required Info, Safety Agreement, and Option Info, Demographics Info,  and show payment instructions.


ryanstrat:
The SCC rules embedded are partial, since the new rules haven't been released yet.

database/migrations/2017_09_20_062736_add_safety_agreement_column.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->timestamp('accept_safety_agreement')->nullable();
```
RossIV: Not that it makes too much difference but for general sanity can you add an `->after('column') to this and put it before the timestamps in this table? I don't have the rest of the schema on hand at the moment to give you a column to put it after.

database/migrations/2017_10_02_062013_add_demographics_columns.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->text('gender')->nullable();
```
RossIV: See above re: column ordering.

resources/assets/js/components/dues/DuesRequiredInfo.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <div class="row">
+    <div class="col-12">
+      <form id="DuesRequiredInfoForm" v-on:submit.prevent="submit">
+        <h3>GT Directory Info</h3>
+        <p>Information obtained via GT Single Sign-On. Update at <a href="https://passport.gatech.edu">Passport</a>.</p>
```
RossIV: Could be misleading to people thinking they can change their name or other information even though not all of it can be changed. Not sure how I'd re-word it though.

resources/assets/js/components/dues/DuesRequiredInfo.vue
```diff
@@ -0,0 +1,171 @@
...
+    },
+    mounted() {
+      var dataUrl = "/api/v1/dues/packages/available";
+      axios.get(dataUrl)
+        .then(response => {
+          this.duesPackages = response.data.dues_packages;
+        })
+        .catch(response => {
+          console.log(response);
+          sweetAlert("Connection Error", "Unable to dues packages. Check your internet connection or try refreshing the page.", "error");
```
RossIV: "Unable to dues packages" I think you a word here.

resources/views/dues/payDues.blade.php
```diff
@@ -0,0 +1,14 @@
+@extends('layouts/app')
+
+@section('title')
+Dues | {{ env('APP_NAME') }}
```
RossIV: In case we do ever end up doing config caching, this should get changed to use the `config()`  accessor for values rather than `env()` directly.

resources/assets/js/components/dues/DuesRequiredInfo.vue
ryanstrat: I think the goal here is to communicate that it cannot be changed within Apiary

ryanstrat:
Changes made, merging PR

PR #175, opened 2017-10-03
Create API endpoints for managing Roles/Permissions

### LINKED ISSUE #170: Create API endpoints for managing Roles/Permissions
Need endpoints for:
- Assigning/removing roles to/from users
- Assigning/removing permissions to/from users/roles
- CRUD roles/permissions

RossIV:
Closes #170


app/Http/Controllers/PermissionController.php
```diff
@@ -0,0 +1,112 @@
...
+        $permission->name = $name;
+        $permission->save();
+
+        $roles = $request->input('roles');
+        if (!empty($roles)) {
+            foreach ($roles as $role) {
+                $dbRole = Role::where('id', '=', $role)->first();
+                if ($dbRole) {
+                    $dbPerm = Permission::find($permission->id);
+                    $dbRole->givePermissionTo($dbPerm);
```
ryanstrat: Is there a reason you go through the effort of finding the `permission` object? Per the documentation, the `givePermissionTo()` can accept a string or a `Spatie\Permission\Models\Permission` object.

app/Http/Controllers/PermissionController.php
```diff
@@ -0,0 +1,112 @@
...
+
+        $name = $request->input('name');
+        $permission = new Permission();
+        $permission->name = $name;
+        $permission->save();
+
+        $roles = $request->input('roles');
+        if (!empty($roles)) {
+            foreach ($roles as $role) {
+                $dbRole = Role::where('id', '=', $role)->first();
```
ryanstrat: I would write this with the `Role::findByName('name');` method.

app/Http/Controllers/PermissionController.php
```diff
@@ -0,0 +1,112 @@
...
+                if ($dbRole) {
+                    $dbPerm = Permission::find($permission->id);
+                    $dbRole->givePermissionTo($dbPerm);
+                } else {
+                    return response()->json(['status' => 'error', 'message' => "Role $role not found."], 422);
+                }
+            }
+        }
+        
+        $dbPermission = Permission::find($permission->id);
```
ryanstrat: permissions also has a `findByName()` method.

app/Http/Controllers/PermissionController.php
```diff
@@ -0,0 +1,112 @@
...
+        return response()->json(['status' => 'success', 'permission' => $dbPermission]);
+    }
+
+    /**
+     * Remove the specified resource from storage.
+     *
+     * @param  int  $id
+     * @return \Illuminate\Http\Response
+     */
+    public function destroy($id)
```
ryanstrat: I don't see the $id of the permission ever being a useful way to identify it externally. Would prefer to use the name and to use the `findByName() method here and in update.

app/Http/Controllers/RoleController.php
```diff
@@ -0,0 +1,149 @@
...
+        
+        $role = new Role();
+        $role->name = $request->input('name');
+        $role->save();
+        
+        if ($request->has('permissions')) {
+            foreach ($request->input('permissions') as $permission) {
+                $dbPerm = Permission::find($permission);
+                if ($dbPerm) {
+                    $role->givePermissionTo($dbPerm);
```
ryanstrat: `$role->givePermissionTo($dbPerm)` can accept a string, a `\Spatie\Permission\Models\Role object` or an `\Illuminate\Support\Collection` object, so we can get rid of the `foreach`

app/Http/Controllers/RoleController.php
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Display the specified resource.
+     *
+     * @param  int  $id
+     * @return \Illuminate\Http\Response
+     */
+    public function show($id)
+    {
+        $role = Role::where('id', $id)->with('permissions')->first();
```
ryanstrat: Change to looking up by name, not id

app/Http/Controllers/RoleController.php
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Update the specified resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  int  $id
+     * @return \Illuminate\Http\Response
+     */
+    public function update(Request $request, $id)
+    {
+        $role = Role::find($id);
```
ryanstrat: name, not id.

app/Http/Controllers/PermissionController.php
RossIV: This pulls from the permission that was just created, no need for a name there since we already have the `id`.

app/Http/Controllers/RoleController.php
RossIV: The foreach was deliberately used to allow for returning an error message that's actually useful to the client in case they specify an incorrect permission name. I can remove it, but your error messages may suck as a result.

PR #176, opened 2017-10-03
Accept Dues UI

### LINKED ISSUE #148: Admin-facing dues workflow
* [ ] Display pending payments
* [ ] Accept a payment
  * Show guidance on accepting checks
  * Trigger receipts to payee
  * Mark user active

ryanstrat:
closes #148 

This PR encompasses the admin table that allows users to view pending dues transactions and mark them as paid.

### DIFF
--- a/app/DuesTransaction.php
+++ b/app/DuesTransaction.php
@@ -56,10 +56,10 @@ class DuesTransaction extends Model
      */
     public function getStatusAttribute()
     {
-        if ($this->payment->first() != null) {
-            return "Paid";
+        if ($this->payment->count() > 0) {
+            return "paid";
         } else {
-            return "Pending";
+            return "pending";
         }
     }
 
--- a/app/DuesTransaction.php
+++ b/app/DuesTransaction.php
@@ -7,11 +7,22 @@ use Illuminate\Database\Eloquent\Model;
 class DuesTransaction extends Model
 {
     /**
+     * The accessors to append to the model's array form.
+     *
+     * @var array
+     */
+    protected $appends = ['status'];
+    
+
+    /**
      * The attributes that aren't mass assignable.
      *
      * @var array
      */
-    protected $guarded = ['id'];
+    protected $guarded = [
+        'id',
+        'status'
+    ];
 
     /**
      * Get the Payment associated with the DuesTransaction model.
@@ -35,6 +46,21 @@ class DuesTransaction extends Model
     public function user()
     {
         return $this->belongsTo('App\User');
+    }
+
+
+    /**
+     * Get the status flag for the Transaction.
+     *
+     * @return bool
+     */
+    public function getStatusAttribute()
+    {
+        if ($this->payment->first() != null) {
+            return "Paid";
+        } else {
+            return "Pending";
+        }
     }
 
     /**
--- a/app/Http/Controllers/PaymentController.php
+++ b/app/Http/Controllers/PaymentController.php
@@ -35,10 +35,19 @@ class PaymentController extends Controller
      */
     public function store(Request $request)
     {
+        $currentUser = auth()->user();
+
+        if (is_null($request->recorded_by) ||
+            $currentUser->cant('update-payments')) {
+            $request['recorded_by'] = $currentUser->id;
+        }
+
         $this->validate($request, [
             'amount' => 'required|numeric',
             'method' => 'required|string',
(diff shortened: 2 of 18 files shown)


app/Http/Controllers/PaymentController.php
```diff
@@ -35,10 +35,19 @@ public function index()
      */
     public function store(Request $request)
     {
+        $currentUser = auth()->user();
+
+        if (is_null($request->recorded_by) ||
```
RossIV: Consider using `!$request->has('recorded_by')` here. As far as I can tell it accomplishes the same thing you're doing here, it's just a more Laravel-esque way of checking.

resources/views/layouts/app.blade.php
```diff
@@ -29,9 +29,11 @@
                 <a class="dropdown-item" href="/admin/users">Users</a>
                 <a class="dropdown-item" href="/admin/events">Events</a>
                 <a class="dropdown-item" href="/admin/faset">Faset</a>
+                <a class="dropdown-item" href="/admin/dues/pending">Accept Dues</a>
```
RossIV: This navigation entry name could potentially be confusing. I get why you named it this, but let's discuss potential alternatives.

resources/assets/js/components/AcceptPayment.vue
```diff
@@ -0,0 +1,170 @@
...
+        payment: {
+          payable_id: this.transactionId,
+          payable_type: "App\\" + this.transactionType,
+          method: "",
+          amount: null
+        },
+        paymentMethods: [
+          {value: "cash", text: "Cash"},
+          {value: "check", text: "Check"},
+          {value: "swipe", text: "Swiped Card"},
```
RossIV: Why are we calling this 'swipe'/'Swiped Card' rather than 'Card' or 'Credit/Debit Card'?

resources/assets/js/components/AcceptPayment.vue
```diff
@@ -0,0 +1,170 @@
...
+          return parseFloat(this.amount);
+        }
+      }
+    },
+    validations: {
+      payment: {
+        method: {
+          required
+        },
+        amount: {
```
RossIV: I _think_ this should be checking for discrepancies in expected vs paid amount if I'm reading this right. If so, it's not working. I recorded a $99 payment for a $10 package.

resources/assets/js/components/dues/DuesTransaction.vue
```diff
@@ -0,0 +1,79 @@
+<template>
```
RossIV: Should this view, after accepting a payment, show any data about the payment on the page? I think that would be useful.

app/DuesTransaction.php
```diff
@@ -37,6 +48,21 @@ public function user()
...
+
+    /**
+     * Get the status flag for the Transaction.
+     *
+     * @return bool
+     */
+    public function getStatusAttribute()
+    {
+        if ($this->payment->first() != null) {
+            return "Paid";
```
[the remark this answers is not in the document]
ryanstrat: I like the idea of using `count()`. The capitalization is so that I don't have to write a render function on the frontend, but that's a bad justification.

resources/views/layouts/app.blade.php
ryanstrat: I'm mulling over a whole reorg of the Navbar. Let's have a separate conversation.

resources/assets/js/components/AcceptPayment.vue
ryanstrat: Because SquareCash is also Debit Card. If we integrate an online cc processor in the future, there may be conflicts. I felt that swiped card was very clear.

resources/assets/js/components/AcceptPayment.vue
ryanstrat: I forgot the code that prevents submit if it's wrong. The validation is working as intended.

resources/assets/js/components/dues/DuesTransaction.vue
ryanstrat: I'm thinking of the use case here where you have 1 person accepting dues from 5 or 10 people in a meeting. A confirmation page will only slow them down.

RossIV: Nah not a confirmation page. I mean on the DuesTransaction page itself when you look at it at any time, assuming the person has the appropriate permissions.

PR #179, opened 2017-10-04
Add Non-Member Role and other fun CAS things

### LINKED ISSUE #4: Implement CAS Login
Implement CAS authentication as middleware

RossIV:
Closes #4


database/seeds/DatabaseSeeder.php
```diff
@@ -11,6 +11,6 @@ class DatabaseSeeder extends Seeder
      */
     public function run()
     {
-        $this->call(RolesAndPermissionsSeeder::class);
+        //
```
ryanstrat: Can we remove empty method?

RossIV: That's there out of the box. I'd rather leave it in case we do want to have more seeds in the future. No harm there by leaving it.

Issue #203, opened 2017-10-18
Prevent users from changing any attributes retrieved from CAS


kberzinch:
The users API should disallow changes to attributes that we receive from CAS, unless the user is an administrator.

kberzinch:
Closed in #207

PR #207, opened 2017-10-20
Prevent users from changing any attributes retrieved from CAS

RossIV:
#203 Adds uid, gtid, first_name, last_name, gt_email to guarded array… to prevent overriding values provided by CAS


kberzinch:
Closes #203

Issue #44, opened 2017-07-01
Implement attendance collection for team meetings


kberzinch:
A kiosk-type page needs to be created to accept BuzzCard swipes for team meetings. Users should be able to select their team in a touch-friendly UI before swiping.

This issue does not include attendance collection for events, such as General Interest.

Magswipe readers will be configured to enter the 9-digit GTID number then press Enter.

RossIV:
#287 could be useful for this, depending on how the front-end for teams goes.

PR #213, opened 2017-11-04
Attendance API

RossIV:
Back-end for #44


app/Http/Controllers/AttendanceController.php
```diff
@@ -0,0 +1,130 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function show(Request $request, $id)
+    {
+        $user = auth()->user();
+        $att = Attendance::find($id);
+        if ($att && ($att->gtid == $user->gtid || $user->can('read-attendance'))) {
+            return response()->json(['status' => 'success', 'attendance' => $att]);
+        } else {
+            return response()->json(['status' => 'error', 'message' => 'Attendance not found.'], 404);
```
ryanstrat: Good job on preventing information leakage through "Permission denied" error messages.

app/Http/Controllers/AttendanceController.php
```diff
@@ -0,0 +1,130 @@
...
+        return response()->json(['status' => 'success', 'attendance' => $attendance]);
+    }
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
```
ryanstrat: The store function is going to return a validation error if an attendance swipe is recorded twice. This is not desirable as in the event the physical card swipers lose the HTTP response, they may re-upload a swipe. In this case, the "additional swipe" should be ignored, in favor of the existing one.

The API should return a 200 (OK) in this case instead of 201 (Created).

app/Http/Controllers/AttendanceController.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        $this->validate($request, [
+            'attendable_type' => 'required|string',
+            'attendable_id' => 'required|numeric',
+            'gtid' => 'numeric|max:9|exists:users',
+            'source' => 'required|string',
+            'recorded_by' => 'required|numeric|exists:users,id'
+        ]);
```
ryanstrat: `created_at` should be accepted from the client-side, since the hardware card swipers may support an offline mode in which the data is uploaded to the API after the event is over.

app/Http/Controllers/AttendanceController.php
RossIV: In testing, this did not result in a validation error. I'm able to submit as many swipes as I'd like for the same event for the same GTID. I will add in de-deduplication and change the response code as requested.

RossIV:
Requested changes have been made.

Issue #217, opened 2017-11-18
Accept card transactions online


kberzinch:
Integrate with a card payments processor to accept card transactions fully online for dues (expand to events once those are live.)

Square is recommended as we already have an established business relationship with them and they support in-person payments.

ryanstrat:
I think this got merged in #218

kberzinch:
The callback from Square isn't implemented yet, so the payment will clear but we won't get that information directly yet.

PR #218, opened 2017-11-20
Add Square Integration for Payment Processing

RossIV:


resources/assets/js/components/AcceptPayment.vue
```diff
@@ -69,6 +69,13 @@
                 <li>Follow the prompts to complete the transaction.</li>
                 <li>It is not required to provide a paper receipt for swiped card transactions.</li>
               </ol>
+            </template>
+              <template v-else-if="payment.method == 'square'">
+              <p class="font-weight-bold">Only use this payment method if you were specifically told to do so.</p>
+              <ul>
+                  <li>Square transactions incur a $3 processing fee. Add $3 to payment amount.</li>
+                  <li>This is <em>not</em> the same as SquareCash.</li>
```
ryanstrat: Not a fan of this copy. If this item is specifically being used only to clean up other system errors, then it should state that in the copy.

kberzinch: This would only be displayed to the treasurer/other admins but I agree it should be rephrased.

kberzinch:
Closes #217

PR #224, opened 2017-11-22
Add User API Token Generation, Squash some bugs

RossIV:


app/Http/Controllers/UserController.php
```diff
@@ -74,9 +85,9 @@ public function store(Request $request)
                 $user->assignRole($requestedRole);
             }
         }
-
+
```
kberzinch: Trailing whitespace

app/Http/Controllers/UserController.php
```diff
@@ -74,9 +85,9 @@ public function store(Request $request)
                 $user->assignRole($requestedRole);
             }
         }
-
+        
         if (is_numeric($user->id)) {
-            $dbUser = User::findOrFail($user->id);
+            $dbUser = User::findOrFail($user->id)->makeVisible('api_token');
```
kberzinch: Why are you doing this here? It's made visible in line 117 after permission checks.

RossIV: Those are two separate functions?

kberzinch: this is a good point

Issue #87, opened 2017-07-21
Add a 404 page if web route not found


ryanstrat:

kberzinch:
Closed in #257

PR #257, opened 2018-01-15
Improve Error/Exception Handling

RossIV:
- Add Bugsnag reporting to all caught exceptions
- Display custom error page instead of Whoops


RossIV:
Closes #87 too (sorry kyla)

PR #328, opened 2018-05-02
Update Readme

ryanstrat:
This isn't complete, but I wanted to go ahead and open this to discussion and completion.


kberzinch:
Note to self, need to update CONTRIBUTING.md

RossIV:
Regarding the #TODO of DB encryption for production, you can't have a functioning install unless you have the appropriate MySQL keychain extension configured so perhaps that should get included here. Otherwise lgtm.

PR #359, opened 2018-05-15
Made table rows clickable

Holben888:
Added some JQuery to datatables.vue that grabs the row ID from the first column and makes the row a clickable link. Also cleaned up the code in the .vue to reduce repeated code.


ryanstrat:
Now that this is merged, there's something breaking on certain tables. Take a look at https://apiary-test.robojackets.org/admin/swag, for just one example.

PR #360, opened 2018-05-16
Clickable table rows (updated)

Holben888:
A check has been added to determine whether or not to use a default pathname. Now, Datatables.vue either defaults to window.location.pathname, or uses a prop passed in called data-link. This fixes broken links resulting from the previous branch. The hyperlinks in the ID column have been removed to simplify the codebase, and the "pointer" cursor is now applied when hovering over table rows.


resources/assets/js/components/Datatable.vue
```diff
@@ -21,6 +21,10 @@ export default {
     dataUrl: String,
     dataObject: Array,
     dataPath: String,
+    dataLink: {
```
ryanstrat: Inline doc comment should be updated with information on the dataLink prop

PR #363, opened 2018-05-17
Added model factories

Holben888:
Added a few models to the ModelFactory.php, namely Payment, User, and DuesPackage. This should make using "artisan tinker" a lot easier, especially when getting this project up and running for the first time. The command for using them is
factory(App\[MODEL_NAME]::class, [NUM_TO_CREATE])->create()


database/factories/ModelFactory.php
```diff
@@ -21,3 +26,72 @@
...
 });
+
+$factory->define(App\DuesPackage::class, function (Faker\Generator $faker) {
+    return [
+        'id' => $faker->numberBetween($min = 0, $max = 200),
+        'name' => $faker->word,
+        'eligible_for_shirt' => $faker->numberBetween($min = 0, $max = 1),
+        'eligible_for_polo' => $faker->numberBetween($min = 0, $max = 1),
+        'effective_start' => $faker->dateTime(),
+        'effective_end' => $faker->dateTime(),
```
RossIV: `effective_end` needs to be after `effective_start`. Maybe `dateTimeBetween()` will do that with some tweaking?

database/factories/ModelFactory.php
```diff
@@ -21,3 +26,72 @@
...
+        'ethnicity' => $faker->randomElement($array = ['white', 'asian', 'hispanic', 'black', 'native', 'islander', 'none']),
+        'accept_safety_agreement' => $faker->optional()->dateTime,
+    ];
+});
+
+$factory->define(App\Payment::class, function (Faker\Generator $faker) {
+    return [
+        'id' => $faker->numberBetween($min = 0, $max = 200),
+        'payable_id' => $faker->numberBetween($min = 0, $max = 200),
+        'payable_type' => $faker->randomElement($array = ['App\DuesTransaction', 'App\Event']),
```
RossIV: If the `DuesTransaction` or `Event` doesn't exist, this may not be of much use. Is that accounted for when this is used? I've not done much with model factories before.

database/factories/ModelFactory.php
```diff
@@ -21,3 +26,72 @@
...
+        'id' => $faker->numberBetween($min = 0, $max = 200),
+        'uid' => $uid,
+        'gtid' => $faker->numerify('#########'),
+        'slack_id' => null,
+        'gt_email' => $uid.'@gatech.edu',
+        'personal_email' => $faker->safeEmail,
+        'first_name' => $faker->firstName,
+        'middle_name' => $faker->optional()->lastName,
+        'last_name' => $lastName,
+        'preferred_name' => $faker->firstName,
```
RossIV: You can probably leave this out, or make it optional. No need to set preferred to the first name if they're the same.

database/factories/ModelFactory.php
Holben888: I believe I was doing this because an App\Payment requires a DuesTransaction or Event to exist to even be created properly, and I couldn't find a good way to autogenerate a DuesTransaction when generating a Payment. Not sure what to do given that dependency...

Holben888: Resolved by creating a `DatabaseSeeder`

kberzinch:
I'm okay ignoring the currently reported Code Climate and Stickler issues. @RossIV please re-review.

PR #366, opened 2018-05-18
Faset DB Cleanup and Freeform Text on FASET form

### LINKED ISSUE #346: FASET database cleanup

ryanstrat:
Included in this PR:

- Migration to remove the `Faset_Surveys` table
- Modifications to Eloquent Models and the FasetVisit Controller for the above table removal
- Adjustment the the JSON schema used to send back data from the FASET form.
- Modifications to the FasitVisit Controller for compatibility to use the new Faset Form JSON schema
- Addition of an 'Other' freeform text option on the Faset Form

closes #346


resources/views/faset/faset.blade.php
```diff
@@ -67,6 +67,13 @@
               <input type="checkbox" class="custom-control-input" id="heardfrom-vex" name="heardfrom" value="vex">
               <label class="custom-control-label" for="heardfrom-vex">Vex Event</label>
             </div>
+            <div class="custom-control custom-checkbox">
+              <input type="checkbox" class="custom-control-input" id="heardfrom-other" name="heardfrom" value="other">
```
kberzinch: Make the Other box only required when the box is checked, otherwise it should be disabled. Here's a helpful snippet from the prototype FASET form:

https://github.com/RoboJackets/apiary/blob/a27231fda6f994e08669748c0b034f8315058e40/resources/views/faset.blade.php#L52-L56

PR #367, opened 2018-05-19
Added privacy policy

### LINKED ISSUE #66: Write Privacy Policy
Given the amount of personal information that this application stores on users, it is important to inform them of how we will and will not use it.

kberzinch:
Adds an unprotected web route to display the privacy policy.

Closes #66.

### DIFF
--- a/resources/views/layouts/footer.blade.php
+++ b/resources/views/layouts/footer.blade.php
@@ -2,7 +2,7 @@
   <div class="container" style="overflow: hidden; height: 100%">
     <div class="d-flex justify-content-between">
       <div>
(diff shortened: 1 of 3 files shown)


kberzinch:
It is not. We could add a footer link.

ryanstrat:
I also support a footer link

ryanstrat:
Regarding Content, it should be added that performing certain actions on the system will require collection of certain information.

Do we need to add anything regarding use of first/third party cookies and/or Google Analytics?

kberzinch:
As of right now, we don't require any data besides what we get from GTED. That will change in the future, of course, and at that point we can update this document.

Your last point is covered in GT's policy and therefore I don't think it warrants inclusion here.

PR #368, opened 2018-05-19
Prevent FASET form navigation if uploads are pending

### LINKED ISSUE #343: More robust FASET upload

kberzinch:
Closes #343


kberzinch:
Only if there are submissions that haven't uploaded (because you're offline or similar.)

PR #369, opened 2018-05-22
Preferred name defaults to legal name on user edit form

### LINKED ISSUE #226: Preferred Name copy
Change input label to "Preferred First Name". Pre-fill with legal first name.

These changes should clarify the purpose of the field.

kberzinch:
This is arguably kind of dumb and should start a discussion about how we store and handle names.

Closes #226

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -48,7 +48,6 @@ class User extends Authenticatable
         'last_name',
         'gt_email',
         'name',
-        'preferred_first_name',
         'gtid',
         'api_token',
         'full_name',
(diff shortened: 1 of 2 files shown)


ryanstrat:
I don't agree with this change.

I like the new storage logic, but I think that we should continue returning null for the preferred name field if one is not set.

kberzinch:
Why?

ryanstrat:
Because the field is null. I don't think we should misrepresent it. Plus, iirc we have a separate field that contains first or preferred,

kberzinch:
That field is `preferred_name`.

ryanstrat:
But basically, the current implementation makes it impossible for the frontend to determine the true state of the database, which is the root of my displeasure.

kberzinch:
I think this is what you want @ryanstrat

kberzinch:
I don't think so. The admin edit page is the same vue component afaik.

PR #374, opened 2018-05-25
Improved validation messages for required fields

### LINKED ISSUE #322: Custom-Radio-Buttons component doesn't support native bootstrap validation styles
Input components allow a user to place feedback in a conditionally visible div directly adjacent to the input field. I'm not sure if the update needed is to bootstrap or to the component, so this needs significant investigation.

```
<div class="invalid-feedback:>
    feedback here
</div>
```

### LINKED ISSUE #353: Edit Profile page won't render if user has no personal email set
On a fresh installation, I couldn't see anything on the Edit Profile page because of a Vue error:
```
app.js:103737 
[Vue warn]: Error in render: "TypeError: Cannot read property 'trim' of undefined"

found in

---> <UserEditForm> at C:\Users\Evan\Documents\GitHub\apiary\resources\assets\js\components\UserEditForm.vue
       <Root>
```

```
TypeError: Cannot read property 'trim' of undefined
    at VueComponent.__webpack_exports__.a (app.js:38138)
    at VueComponent.runRule (app.js:114303)
    at VueComponent.run (app.js:114345)
    at Watcher.get (app.js:106286)
    at Watcher.evaluate (app.js:106393)
    at VueComponent.computedGetter [as run] (app.js:106651)
    at VueComponent.proxy (app.js:114351)
    at Watcher.get (app.js:106286)
    at Watcher.evaluate (app.js:106393)
    at VueComponent.computedGetter [as proxy] (app.js:106651)
```

I then set the `personal_email` attribute for my user using Tinker, and Edit Profile works now.

I suspect the issue is with the notGTEmail validator calling `trim` on a null value (which is apparently the initial value of the personal_email attribute), specifically:
```
!notGTEmail.trim().toLowerCase().endsWith('gatech.edu');
```

evan10s:
The custom radio button input can now show an error message using a similar HTML structure as the other invalid messages.

While I was at it, I looked around and added or fixed up the validation messages in various places, along with the notGTEmail validator bug from #353.

Closes #353, closes #322


resources/assets/js/components/UserEditForm.vue
```diff
@@ -27,29 +27,29 @@
         <h3>Additional Information</h3>
 
         <div class="form-group row">
-          <label for="user-preferredname" class="col-sm-2 col-form-label">Preferred First Name</label>
+          <label for="user-preferredname" class="col-sm-2 col-form-label">Preferred Name</label>
```
kberzinch: These changes clobber a PR I merged recently. This field should be `preferred_first_name`.

kberzinch:
I believe this PR closes the two issues I added to the description; please correct me if not.

resources/assets/js/components/UserEditForm.vue
```diff
@@ -219,8 +224,8 @@ export default {
       personal_email: { email, notGTEmail },
       phone: { maxLength: maxLength(15) },
       preferred_first_name: { alpha },
-      shirt_size: {},
-      polo_size: {},
+      shirt_size: { required },
```
kberzinch: Technically we don't need your shirt size unless you pay dues, so I'm not sure if this validation is per se correct.

evan10s:
The same fields are also in the dues flow iirc, so they could be optional
on the edit profile page.

PR #411, opened 2018-08-28
Fix attendance duplication

JoshOldenburg:
The duplication check only worked for the current day, not when manually
adding attendance from the past.


JoshOldenburg:
Is the intent one record per person per day or one record per person per team per day?

RossIV:
Per person per team per day @JoshOldenburg

kberzinch:
The latter

JoshOldenburg:
In that case, I tested this fix and it works.

app/Http/Controllers/AttendanceController.php
```diff
@@ -53,18 +53,18 @@ public function store(Request $request)
         unset($request['includeName']);
 
         // Variables for comparison below
-        $today = date('Y-m-d');
+        $date = $request->has('created_at') ? $request->input('created_at') : date('Y-m-d');
```
RossIV: This can be simplified to `$date = $request->input('created_at', date('Y-m-d');`

app/Http/Controllers/AttendanceController.php
```diff
@@ -53,18 +53,18 @@ public function store(Request $request)
...
         $gtid = $request->input('gtid');
 
         try {
-            $attTodayQ = Attendance::where($request->all())->whereDate('created_at', $today);
-            $attTodayCount = $attTodayQ->count();
-            if ($attTodayCount > 0) {
-                Log::debug(get_class().": Found a swipe today ($today) for $gtid - ignoring.");
-                $att = $attTodayQ->first();
+            $attExistingQ = Attendance::where($request->all())->whereDate('created_at', $date);
```
RossIV: Have you tested how this behaves when you specify a `created_at` in the request?

JoshOldenburg: Yes, it works with and without `created_at`.

PR #417, opened 2018-08-29
Pre-GI Tweaks

RossIV:
Lots of improvements to GI invite email flow. Didn't get everything done that we wanted, but it's good enough for now and (more importantly) much better than it was before.


app/Http/Controllers/RecruitingCampaignController.php
```diff
@@ -0,0 +1,165 @@
...
+                Log::info(get_class().": Email '$v->recruiting_email' already in the list. Ignoring.'");
+            } else {
+                // Add new recipient
+                $rcr = new RecruitingCampaignRecipient();
+                $rcr->email_address = $v->recruiting_email;
+                $rcr->source = 'recruiting_visit';
+                $rcr->recruiting_visit_id = $v->id;
+                $rcr->recruiting_campaign_id = $rc->id;
+                if ($v->user_id != null) {
+                    $rcr->user_id = $v->user_id;
```
kberzinch: Why do we copy this

RossIV: the real answer now that I think about this more would be to have made that class (recipients) polymorphic and not copy any of the data

but I didn't think about that until now, so that's why that's there

PR #458, opened 2018-09-21
Recruiting in Nova

kberzinch:
Rebuild of main recruiting admin interface. Does not implement form editing as RecruitingResponses do not have primary keys.

Definitely remaining work:
- [x] Add action for sending individual emails

Maybe remaining work:
 - [ ] RecruitingCampaigns and related models?


kberzinch:
good enough for me

Issue #161, opened 2017-09-20
Implement Fractal


RossIV:
Investigate the use of [Fractal](https://fractal.thephpleague.com) for retrieval of Eloquent models and implement if so deemed necessary.

RossIV:
Per discussion in Slack, this is being postponed to a future release.

PR #459, opened 2018-09-21
Fixes #457: Create endpoint for displaying user's teams

### LINKED ISSUE #457: Add API endpoint for getting all teams for a user

Arthelon:
Fixes #457.

Only validates that the requesting user is an admin.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -161,7 +161,8 @@ class UserController extends Controller
      * @param Request $request
      * @return \Illuminate\Http\JsonResponse
      */
-    public function showTeams($id, Request $request) {
+    public function showTeams($id, Request $request)
+    {
         $user = User::findByIdentifier($id)->first();
         if ($user) {
             $requestingUser = $request->user();
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -155,6 +155,30 @@ class UserController extends Controller
     }
 
     /**
+     * Displays the teams the specified user belongs to.
+     *
+     * @param int $id
+     * @param Request $request
+     * @return \Illuminate\Http\JsonResponse
+     */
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -154,6 +154,31 @@ public function show($id, Request $request)
...
+     * @param int $id
+     * @param Request $request
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function showTeams($id, Request $request)
+    {
+        $user = User::findByIdentifier($id)->first();
+        if ($user) {
+            $requestingUser = $request->user();
+            if ($requestingUser->hasRole('admin')) {
```
kberzinch: This should be a check for a permission instead of a role. The permission we need here does not exist yet afaik, but we will need to also update the corresponding team api endpoint.

kberzinch: // @RossIV

RossIV: As discussed in Slack, see `2018_05_27_220626_add_hidden_teams_permission.php` as an example for permissions for this.

routes/api.php
```diff
@@ -43,6 +43,7 @@
     Route::post('attendance/search', 'AttendanceController@search')->name('attendance.search');
     Route::resource('attendance', 'AttendanceController', ['except' => ['create', 'edit']]);
     Route::get('users/search', 'UserController@search');
+    Route::get('users/{id}/teams', 'UserController@showTeams');
```
RossIV: This route needs a name like what's seen on like 43 in the same file

app/Http/Controllers/UserController.php
kberzinch: Add a permission for users to view team membership and we'll use it for #461 once that's a thing. Add the permission to the admin and officer roles.

kberzinch:
Closing in favor of whatever gets implemented for #161

PR #467, opened 2018-09-24
Add name to nova/logout route

### LINKED ISSUE #451: nova.logout route expected but not registered in some envs
Partial fix on @JoshOldenburg's local repository but should revisit

JoshOldenburg:
Nova was failing to load because it could not find the route named nova.logout. Routes can't have multiple names so this gives the name to the redirect. I'm open to better solutions; I couldn't find a more elegant fix.

Closes #451

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -13,7 +13,7 @@
 
 Route::get('nova/logout', function () {
     return redirect('logout');
-});
+})->name('nova.logout');
 
 Route::group(['middleware' => 'auth.cas.force'], function () {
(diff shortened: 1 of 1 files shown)


kberzinch:
Assigning to Ross per Slack:
> I'm about to run away from my computer for a bit but I may play with that later
<sub>[View in Slack](https://robojackets.slack.com/archives/C5A25H2GM/p1537826246000100)</sub>

kberzinch:
Per slack this is the least bad way to do this.

PR #468, opened 2018-09-24
Teams Tweaks

JoshOldenburg:
Filter the kiosk to only show visible and attendable teams, filter the attendance admin page to only show attendable teams, and stop refreshing the team edit page every time it is submitted.

### DIFF
--- a/resources/assets/js/components/attendance/AttendanceKiosk.vue
+++ b/resources/assets/js/components/attendance/AttendanceKiosk.vue
@@ -43,7 +43,9 @@ export default {
           if (rawTeams.length < 1) {
             swal('Bueller...Bueller...', 'No teams found.', 'warning');
           } else {
-            this.teams = response.data.teams.sort(function(a, b) {
+            this.teams = response.data.teams.filter(function(item) {
+              return item.visible;
+            }).sort(function(a, b) {
               return a.name > b.name ? 1 : b.name > a.name ? -1 : 0;
             });
             this.startListening();
--- a/resources/assets/js/components/attendance/AttendanceKiosk.vue
+++ b/resources/assets/js/components/attendance/AttendanceKiosk.vue
@@ -44,7 +44,7 @@ export default {
             swal('Bueller...Bueller...', 'No teams found.', 'warning');
           } else {
             this.teams = response.data.teams.filter(function(item) {
-              return item.visible && item.attendable;
+              return item.attendable;
             }).sort(function(a, b) {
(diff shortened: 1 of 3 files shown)


resources/assets/js/components/teams/TeamEditForm.vue
```diff
@@ -216,7 +216,8 @@ export default {
           this.hasError = false;
           this.feedback = 'Saved!';
           console.log('success');
-          window.location.href = '/admin/teams/' + response.data.team.slug;
+          var newHREF = '/admin/teams/' + response.data.team.slug;
+          if (!window.location.href.endsWith(newHREF)) window.location.href = newHREF;
```
kberzinch: Why is the reload necessary?

resources/assets/js/components/attendance/AttendanceKiosk.vue
```diff
@@ -43,7 +43,9 @@ export default {
           if (rawTeams.length < 1) {
             swal('Bueller...Bueller...', 'No teams found.', 'warning');
           } else {
-            this.teams = response.data.teams.sort(function(a, b) {
+            this.teams = response.data.teams.filter(function(item) {
+              return item.visible && item.attendable;
```
kberzinch: Only check for `attendable` - we may want to have teams that appear on the kiosk but not on the website (training maybe???)

resources/assets/js/components/teams/TeamEditForm.vue
JoshOldenburg: That form is also used for creating the team initially, so it'll redirect from `/admin/teams/create` to `/admin/teams/<slug>`. When editing an existing one it won't redirect now.

kberzinch: Oh derp this is on the edit form, carry on.

resources/assets/js/components/attendance/AttendanceKiosk.vue
kberzinch: Training definitely per bossman

resources/assets/js/components/attendance/AttendanceKiosk.vue
```diff
@@ -43,7 +43,9 @@ export default {
           if (rawTeams.length < 1) {
             swal('Bueller...Bueller...', 'No teams found.', 'warning');
           } else {
-            this.teams = response.data.teams.sort(function(a, b) {
+            this.teams = response.data.teams.filter(function(item) {
+              return item.attendable;
```
kberzinch: Switch back to attendable and visible

Issue #335, opened 2018-05-08
Dues flow flashing checkboxes


kberzinch:
Not sure if this is still an issue, but we should check.

>On the ethnicity dialogue, when I click next all of the check boxes flashed checked before the page proceeded.

kberzinch:
Confirmed on staging today

kberzinch:
Seems to have disappeared

PR #469, opened 2018-09-25
Dues flow back button

### LINKED ISSUE #338: Step back to previous pages in dues flow
>Itd be nice to be able to move back while going through the "pay dues" process. That is, I'm on the pick a payment method page but I'd like to go back and edit my emergency contact's info. Pressing the back button takes me back to the picking a t-shirt size page.

JoshOldenburg:
Fixes #338

Also makes the skip button not submit the form.

There's no back button on the final payment screen, though technically it could have one.

image


app/Http/Controllers/DuesTransactionController.php
```diff
@@ -120,8 +120,8 @@ public function store(Request $request)
         $existingTransaction = DuesTransaction::where('dues_package_id', $request->input('dues_package_id'))
             ->where('user_id', $request->input('user_id'))->first();
         if ($existingTransaction) {
-            return response()->json(['status' => 'error',
-                'message' => 'There is already a pending Dues Transaction for this user', ], 400);
+            $existingTransaction->touch(); //Update the updated_date on the transaction so it takes precedence
```
JoshOldenburg: @RossIV you're okay with this solution for keeping track of the most recently selected transaction?

JoshOldenburg:
It should maintain the previous state now.

JoshOldenburg:
Don't merge this yet, I found an issue with it. On the demographics screen, checking one box causes all of them to be checked.

kberzinch:
That may or may not be a bug with your change - take a look at #335.

app/Http/Controllers/DuesTransactionController.php
```diff
@@ -116,12 +116,23 @@ public function store(Request $request)
...
+        // and replace it with the one currently being requested
+        if ($user->dues->count() > 0) {
+            $existingTransaction = $user->dues->last();
+            $pkgIsActive = $existingTransaction->package->is_active;
+            if ($pkgIsActive) {
+                $hasPayment = $existingTransaction->payment()->exists();
+                if ($hasPayment) {
+                    $paymentAmount = $existingTransaction->payment->sum('amount');
+                    $paidTotal = ($paymentAmount >= $existingTransaction->getPayableAmount());
+                    if (! $paidTotal) {
```
RossIV: If they paid anything, IMO we shouldn't be deleting it because that's money. I get that it's only soft delete, but still.

app/Http/Controllers/PaymentController.php
```diff
@@ -111,9 +111,9 @@ public function storeUser(Request $request)
         } else {
             //Assuming DuesTransaction for now
 
-            //Find DuesTransactions without payment attempts
+            //Find the most recent DuesTransaction without a payment attempt
             $transactWithoutPmt = DuesTransaction::doesntHave('payment')
-                ->where('user_id', $user->id)->first();
+                ->where('user_id', $user->id)->get()->sortBy('updated_at')->last();
```
RossIV: By calling `get()` then `last()` you're first getting all transactions for the user without payment and then getting the last. I think it would be more efficient to take out the `get()` but I haven't tested that.

app/Http/Controllers/DuesTransactionController.php
JoshOldenburg: I agree. In what case will the payment amount be less than the total?

kberzinch: Can we leave partially paid transactions?

JoshOldenburg: It is leaving partially paid transactions now.

kberzinch: Ok, this seems fine.

PR #471, opened 2018-09-26
Add team joining to the dues flow

### LINKED ISSUE #438: Add team affiliation update to dues flow

JoshOldenburg:
image

There's no back button until #469 is merged, and there's a somewhat hacky way of getting the teams object for the current user until #459 is merged.

Teams that aren't self-serviceable are not shown. Any suggestions for a better wording of the header?

Fixes #438.


kberzinch:
I was definitely imagining more of a "here are the teams you're in, stop now and adjust on the teams page if wrong". That may be more maintainable.

Can't you reuse the cards component from the teams page? Or is that not done in Vue?

JoshOldenburg:
The card component was not Vue, but I ported it to Vue in this PR. The teams page now uses that component. The Vue component is the individual card.

Could the same UI with a different message be used for what you're suggesting? Something like "Team Membership" subheading "This is the list of teams we know you're on. If this isn't accurate, please update it here." and remove the skip button?

kberzinch:
Ah, I didn't catch that. If you've made those separate components then yeah that would work.

JoshOldenburg:
image

Better?

resources/assets/js/components/teams/TeamCard.vue
```diff
@@ -0,0 +1,34 @@
...
+                <p class="card-text">{{ team.description }}</p>
+                <template v-if="team.mailing_list_name || team.slack_channel_id">
+                    <hr/>
+                    <p class="card-text">
+                    <template v-if="team.mailing_list_name">
+                        Mailing List: <a :href="'https://lists.gatech.edu/sympa/info/' + team.mailing_list_name" target="_blank">{{ team.mailing_list_name }}</a>
+                        <br/>
+                    </template>
+                    <template v-if="team.slack_channel_id">
+                        Slack: <a :href="'https://slack.com/app_redirect?team=T033JPZLT&channel=' + team.slack_channel_id" target="_blank">{{ team.slack_channel_name }}</a>
```
kberzinch: Add a `#` here as on prod

resources/assets/js/components/dues/DuesSequence.vue
```diff
@@ -17,6 +17,27 @@
...
 
+    <div v-if="currentStepName == 'join-teams'">
+      <div class="row">
+        <div class="col-12">
+          <h3>Team Membership</h3>
+        </div>
+      </div>
+      <div class="row">
+        <div class="col-12">
+          <p>This is the list of teams we know you're on. If this isn't accurate, please update it here.</p>
```
kberzinch: Add a note about the standalone teams page here.

JoshOldenburg:
Done. This should probably wait until #469 is merged so I can add a back button to the page.

JoshOldenburg:
Back button:
image

PR #472, opened 2018-09-28
Prettier team attendance interface

JoshOldenburg:
As discussed in Slack, this is a nicer interface for team attendance that can accept raw BuzzCard swipes, particularly for trainings. The original interface still works as before, with just an additional button to get the modal.

Before selecting a team, there's a tooltip and the button is disabled:
image

The button brings up the same modal as the event admin page:
image


JoshOldenburg:
Now there's a dropdown:
image

I briefly considered doing cards like /teams, or for admins showing hidden teams and "record swipes" buttons on /teams, but decided this was much simpler and would work just as well.

Issue #434, opened 2018-09-09
Proper RBAC for Nova


kberzinch:

kberzinch:
The roles and permissions admin UIs may or may not have any useful options to lock them down.

I didn't like those to begin with so if we need to rebuild them then we should.

JoshOldenburg:
What do you not like about them? I.e. what would you change if we reimplemented it?

kberzinch:
It's just ugly

PR #501, opened 2018-10-10
Add more Nova resources

### LINKED ISSUE #473: Attendance admin in Nova
I don't anticipate this actually being particularly useful but it would be nice to show on teams and events and users.

### LINKED ISSUE #474: Event admin in Nova
Related to #473

JoshOldenburg:
Fixes #474, fixes #473. Starts #434 (only has policies for attendance and RSVPs).

Attendance resource page:
screen shot 2018-10-10 at 5 10 14 pm

Events resource page:
image

Event detail page has RSVPs and attendance:
image

RSVP detail page: (not linked in the sidebar, only from events)
image

Teams and user pages also have attendance on them. Attendance and RSVPs can't be created or edited in Nova.

### DIFF
--- a/app/Nova/Attendance.php
+++ b/app/Nova/Attendance.php
@@ -3,16 +3,11 @@
 namespace App\Nova;
 
 use Laravel\Nova\Panel;
-use Laravel\Nova\Fields\ID;
 use Illuminate\Http\Request;
 use Laravel\Nova\Fields\Text;
-use Laravel\Nova\Fields\BelongsTo;
-use Laravel\Nova\Fields\Boolean;
-use Laravel\Nova\Fields\HasMany;
-use Laravel\Nova\Fields\DateTime;
-use Laravel\Nova\Fields\Textarea;
-use Laravel\Nova\Fields\Currency;
 use Laravel\Nova\Fields\MorphTo;
+use Laravel\Nova\Fields\DateTime;
+use Laravel\Nova\Fields\BelongsTo;
 
 class Attendance extends Resource
 {
--- a/app/Nova/Attendance.php
+++ b/app/Nova/Attendance.php
@@ -1,0 +1,162 @@
+<?php
+
+namespace App\Nova;
+
+use Laravel\Nova\Panel;
+use Laravel\Nova\Fields\ID;
+use Illuminate\Http\Request;
+use Laravel\Nova\Fields\Text;
+use Laravel\Nova\Fields\BelongsTo;
+use Laravel\Nova\Fields\Boolean;
+use Laravel\Nova\Fields\HasMany;
+use Laravel\Nova\Fields\DateTime;
+use Laravel\Nova\Fields\Textarea;
+use Laravel\Nova\Fields\Currency;
+use Laravel\Nova\Fields\MorphTo;
+
+class Attendance extends Resource
+{
+    /**
+     * The model the resource corresponds to.
+     *
+     * @var string
+     */
+    public static $model = 'App\Attendance';
+
+    /**
+     * Get the displayble label of the resource.
+     *
+     * @return string
+     */
+    public static function label()
+    {
+        return 'Attendance';
+    }
+
+    /**
+     * Get the displayble singular label of the resource.
+     *
+     * @return string
+     */
+    public static function singularLabel()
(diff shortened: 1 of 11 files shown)


app/Nova/Attendance.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    protected function basicFields()
+    {
+        return [
+            Text::make('GTID')
+                ->sortable()
+                ->rules('required', 'max:255'),
+
+            (new BelongsTo('User', 'attendee', 'App\Nova\User'))
```
kberzinch: Use `BelongsTo::make` here.

app/Nova/Attendance.php
```diff
@@ -0,0 +1,162 @@
...
+                ->sortable(),
+
+            (new MorphTo('Attended', 'attendable'))
```
kberzinch: Use `MorphTo::make` here.

app/Nova/Attendance.php
```diff
@@ -0,0 +1,162 @@
...
+                ->sortable()
+                ->types([
+                    Event::class,
+                    Team::class,
+                ]),
+
+            (new BelongsTo('Recorded By', 'recorded', 'App\Nova\User'))
```
kberzinch: Use `BelongsTo::make` here.

app/Nova/Event.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    protected function basicFields()
+    {
+        return [
+            Text::make('Event Name', 'name')
+                ->sortable()
+                ->rules('required', 'max:255'),
+
+            (new BelongsTo('Organizer', 'organizer', 'App\Nova\User'))
```
kberzinch: Use `BelongsTo::make` here.

app/Providers/AuthServiceProvider.php
```diff
@@ -0,0 +1,33 @@
...
+namespace App\Providers;
+
+use App\Attendance;
+use App\Rsvp;
+use App\Policies\AttendancePolicy;
+use App\Policies\RsvpPolicy;
+use Illuminate\Support\Facades\Gate;
+use Illuminate\Foundation\Support\Providers\AuthServiceProvider as ServiceProvider;
+
+class AuthServiceProvider extends ServiceProvider
```
kberzinch: Thoughts on renaming this to something more specific? In this app it will only ever be used for policies.

app/Nova/Attendance.php
JoshOldenburg: I can't use it here and still have the interface refer to it as "Recorded By".

app/Providers/AuthServiceProvider.php
JoshOldenburg: I followed the Laravel suggestion for the name, but this could just be combined into the NovaServiceProvider if you'd prefer

app/Nova/Event.php
JoshOldenburg: Same as above

kberzinch:
I may be misremembering something but I _think_ you can do attribute-level policies. We should do that for GTIDs and API tokens.

app/Nova/Attendance.php
kberzinch: Interesting.

JoshOldenburg:
That is possible., but those permissions don't exist right now. I was planning to add those across Nova once the permissions from Ross's spreadsheet are created.

app/Providers/AuthServiceProvider.php
kberzinch: Hmmm. I think that would be cleaner.

app/Nova/Attendance.php
JoshOldenburg: The make method doesn't pass those through to the constructor as far as I can tell.

kberzinch:
Gotcha.

app/Providers/AuthServiceProvider.php
JoshOldenburg: On second look that's not an option, it has to inherit from the AuthServiceProvider superclass.

kberzinch: That's fine then

Issue #461, opened 2018-09-21
Team members API should be restricted maybe


kberzinch:
Blocking #459 #457

kberzinch:
Implemented in #505

PR #505, opened 2018-10-11
Implement API Resource Functionality

### LINKED ISSUE #161: Implement Fractal

RossIV:
Implements Laravel's API Resources for finer-grained control of attribute release and for dynamic inclusion of related models in queries

This needs a lot of testing to see what broke as a result of the API changes.

Closes #161 
Maybe #461? Damn Kristaps and his super generic issues
Related to #457


RossIV:
>/api/v1/events?include=organizer throws an exception for me. Am I calling that with the wrong parameter?

@JoshOldenburg That's fixed, was a typo.

RossIV:
>Also, was changing the format of the dates returned in JSON intentional? 

@JoshOldenburg That wasn't intentional directly, though I think that comes with the resource functionality. I can see if it can be disabled if we want to disable it.

RossIV:
Things that this has broken, some of which we've deemed to be acceptable casualties:

1. GTID display in any datatable
2. Emergency Contact Info in Admin -> Users -> Edit User Data (You can't view it but editing + saving works.)
3. All timestamps show up as either empty fields or `[object Object]` in tables

There could be more and there probably is. That's not an exhaustive list.

PR #512, opened 2018-10-13
Add "fast-read" kiosk checkbox

### LINKED ISSUE #504: Fast read option for attendance kiosk
This came up in core - basically add an option to "stick" a team so people can swipe continuously.

Suggested as prereq for a useful contactless solution.

Arthelon:
Fixes #504 

This PR implements a "fast-read" option for the attendance kiosk by adding a checkbox that when checked will not reset the stored team ID upon successful swipes. 

This PR also improves the attendance UI by adding loaders during submit requests and prevents multiple swipes from being processed at the same time. 

**NOTE:** When a fast-read swipe is successful, the loading indicator disappears. (If no error modal appears then the swipe is successful)

### DIFF
--- a/resources/assets/js/components/attendance/AttendanceKiosk.vue
+++ b/resources/assets/js/components/attendance/AttendanceKiosk.vue
@@ -182,12 +182,10 @@ export default {
           showCancelButton: true,
           showConfirmButton: false,
           type: 'warning',
-        }).then(result => {
-          if (!result.value) {
+          onClose: () => {
             self.clearFields();
-            swal.close();
           }
-        });
+        })
       } else {
         swal.close();
         console.log('unknown cardData: ' + cardData);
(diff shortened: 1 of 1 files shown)


Arthelon:
I think the size of the checkbox is fine since clicking on the text label will trigger it. I made the font-size of the label a bit bigger and separated it more from the other modal elements to make it a bit more accessible.

Also added the existing swal popup to the "fast-read" flow.

kberzinch:
Yeah, but does your average RoboJackets member know that clicking the text works? I don't know about that.

Arthelon:
They don't have to know about clicking on the label. The point is that when they try to click on the checkbox it's going to work most of the time.

kberzinch:
Thoughts on cutting the confirmation popup to 1 second instead of 2?

Arthelon:
that might be better when fast-read is enabled.

kberzinch:
I think it would be fine even if it was disabled. You really just need to see the green checkmark.

kberzinch:
Can you adjust that? Then I think this is good to merge.

PR #539, opened 2018-11-01
Tweaks to attendance report

kberzinch:


app/Http/Controllers/AttendanceController.php
```diff
@@ -205,7 +205,7 @@ public function statistics(Request $request)
         // so it can be sorted more easily; the number is trimmed out in the map method
         $attendanceByDay = Attendance::whereBetween('created_at', [$startDay, $endDay])
             ->where('attendable_type', 'App\Team')
-            ->selectRaw('date_format(created_at, \'%w%W\') as day, count(gtid) as aggregate')
+            ->selectRaw('date_format(created_at, \'%w%W\') as day, count(distinct gtid) as aggregate')
```
JoshOldenburg: This intentionally didn't include `distinct` because it will only count an individual once per Mon/Tue/Wed/etc (i.e. only one week). To do the average it needs to include the count from every week.

kberzinch: Gotcha

PR #574, opened 2019-02-26
Initial access control changes

### LINKED ISSUE #433: Separate "active" vs "paid" start/end dues dates
We are going to be using Apiary as a system of record for system access authorization very soon. As part of this process, we need to distinguish when a user is no longer active.

Active in this context means something different from active in the context of "do you need to pay dues this semester". In the former case, we need users to be "active" through the dues "deadline" so that they do not lose access to systems. In the latter case, we need users to be "inactive" as early as reasonable so that they are able to pay dues for the new semester.

My proposed solution to this is to have separate start/end dates for when a user is considered active for dues status vs active for access status.

### LINKED ISSUE #541: Add Project Manager field to teams
Used for restricting team membership management and editing a team

### LINKED ISSUE #577: Delete old officer roles
smh @JoshOldenburg

kberzinch:
Adds concept of "Access Active" to DP, DT, users.

Closes #433
Closes #541
Closes #577 

### Todo
- [x] Restrict team attach/detach to project manager of the team or `admin`s

### ~Out of scope~
~Pushing changes to JEDI~


kberzinch:
Outstanding questions from Slack:
 - How do we record temporary access overrides? i.e., new people that may need access to systems before committing to pay dues. Special team?
 - How do we record permanent access overrides? i.e. alumni who are still contributing but no longer pay dues. Flag on user?

app/Nova/User.php
```diff
@@ -272,6 +275,12 @@ public function lenses(Request $request)
     public function actions(Request $request)
     {
         return [
+            (new Actions\SyncAccess)
+                ->canSee(function ($request) {
+                    return true;
+                })->canRun(function ($request, $user) {
+                    return true;
```
JoshOldenburg: Is this meant to be available to everyone? I know in theory it is harmless to run at arbitrary times, but it could be potentially confusing to users.

app/Nova/User.php
```diff
@@ -272,6 +275,12 @@ public function lenses(Request $request)
     public function actions(Request $request)
     {
         return [
+            (new Actions\SyncAccess)
+                ->canSee(function ($request) {
+                    return true;
```
JoshOldenburg: This shouldn't be visible if Jedi integration isn't configured.

app/User.php
```diff
@@ -289,4 +321,34 @@ public function scopeInactive($query)
...
+     * Active: Has paid dues for a currently ongoing term
+     *         or, has a non-zero payment for an active DuesPackage.
+     * @param \Illuminate\Database\Eloquent\Builder $query
+     * @param mixed $type
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopeAccessActive($query)
+    {
+        return $query->whereHas('dues', function ($q) {
+            $q->paid()->accessCurrent();
```
JoshOldenburg: This doesn't check override teams

app/User.php
```diff
@@ -289,4 +321,34 @@ public function scopeInactive($query)
...
+     * Active: Has paid dues for a currently ongoing term
+     *         or, has a non-zero payment for an active DuesPackage.
+     * @param \Illuminate\Database\Eloquent\Builder $query
+     * @param mixed $type
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopeAccessInactive($query)
+    {
+        return $query->whereDoesntHave('dues', function ($q) {
+            $q->paid()->accessCurrent();
```
JoshOldenburg: This doesn't check override teams

app/Nova/User.php
kberzinch: It is intentionally visible to anyone with Nova access.

I guess only admins would need to run it though. Will fix.

app/User.php
kberzinch: That's in the category of SQL I could probably write but really don't want to.

This currently isn't used anywhere except in the access_active attribute getter which already checks teams.

app/User.php
kberzinch: Same as above.

app/Nova/Actions/SyncAccess.php
```diff
@@ -0,0 +1,69 @@
...
+     * @return mixed
+     */
+    public function handle(ActionFields $fields, Collection $models)
+    {
+        // This whole function is going to get factored out later
+        // Also it will be a queued job I guess?
+        foreach ($models as $user) {
+            $send = [];
+            $send['uid'] = $user->uid;
+            $send['preferred_first_name'] = $user->preferred_first_name;
```
RossIV: Why not just name it `$send['first_name']` - Does Jedi also have different fields for first/preferred first?

app/Nova/Actions/SyncAccess.php
```diff
@@ -0,0 +1,69 @@
...
+                        'User-Agent' => 'Apiary on '.config('app.url'),
+                        'Accept' => 'application/json',
+                        'Authorization' => 'Bearer '.config('jedi.token'),
+                    ],
+                ]
+            );
+
+            $response = $client->request('POST', config('jedi.endpoint'), ['json' => $send]);
+
+            if (200 !== $response->getStatusCode()) {
```
RossIV: should there be an else here?

app/Nova/Team.php
```diff
@@ -61,6 +62,10 @@ public function fields(Request $request)
                 ->alwaysShow()
                 ->rules('required'),
 
+            BelongsTo::make('Project Manager', 'projectManager', User::class)
```
RossIV: Do we care about Co-PMs for this use case?

app/Policies/TeamPolicy.php
```diff
@@ -57,7 +57,14 @@ public function create(User $user)
      */
     public function update(User $user, Team $team)
     {
-        return $user->can('update-teams');
+        if ($user->can('update-teams')) {
```
RossIV: multiple separate `if` statements feels weird here rather than if/elif/else

app/Policies/UserPolicy.php
```diff
@@ -122,7 +128,22 @@ public function attachTeam(User $user, User $userResource, Team $team)
...
     {
-        return $user->can('update-teams-membership');
+        if ($user->can('update-teams-membership')) {
+            return true;
+        }
+        if ($user->can('update-teams-membership-own') && $user->is($userResource)) {
+            return true;
+        }
+        $user_manages = $user->manages()->get();
+        if (count($user_manages) > 0) {
```
RossIV: I don't think this logic works?

> This method will return the values in the original collection that are not present in the given collection

For example, user manages Outreach and is a member of Outreach and Battlebots. The attach command is called for Battlebots.
```
$user_manages = ['Outreach'];
$target_in = ['Outreach', 'Battlebots'];
$diff = ['Battlebots'];
```
Since there's a difference, it'll allow it which I don't think is the correct behavior since the user doesn't manage Battlebots. That is, unless I have something backwards.

app/User.php
```diff
@@ -240,6 +253,25 @@ public function getIsActiveAttribute()
...
+     * Get the access_active flag for the User.
+     *
+     * @return bool
+     */
+    public function getIsAccessActiveAttribute()
+    {
+        if (self::where('id', $this->id)->accessActive()->count() != 0) {
+            return true;
+        }
+        foreach ($this->teams()->get() as $team) {
```
RossIV: Is there another way that isn't a foreach loop to do this?

database/migrations/2019_02_25_171041_add_jedi_fields.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public function up()
+    {
+        Schema::table('dues_packages', function (Blueprint $table) {
+            $table->timestamp('access_start')->after('effective_end')->useCurrent();
+            $table->timestamp('access_end')->after('access_start')->useCurrent();
+        });
+
+        Schema::table('teams', function (Blueprint $table) {
+            $table->unsignedInteger('project_manager')->nullable()->comment('user_id of the project manager');
```
RossIV: Please make this `project_manager_id` to avoid potential conflict with relationship names on the model.

app/Nova/Actions/SyncAccess.php
kberzinch: I don't think so. Nova marks the action as completed as soon as this function finishes.

app/Nova/Team.php
kberzinch: No.

app/Policies/UserPolicy.php
kberzinch: Battlebots isn't in 'the original collection'

app/User.php
kberzinch: You could probably do it in SQL but it would be gross. I can try it if you want.

RossIV: Instead of fetching the entire model of each team, can you just get an array of team `id`s and do an `array_intersect()` with the access teams from config?

kberzinch:
Notes from Core/Slack
 - There is a desire to give new members access to services (specifically Vault) before they pay dues.
   - I don't really want to do service-specific permissions for the sake of our collective sanity, so we're going to treat it like a normal override.
   - This system can be admin-entered.
 - Proposed implementation is a datetime field on the user that represents the time at which their access override expires. We can create an action to set this value for non-admins to a sane value (i.e. dues deadline). This can also be used for "permanent" access overrides, you would just need to do annual renewals or similar.

cc @RossIV for opinions

app/User.php
kberzinch: Resolving this thread as we're no longer using teams for access control

app/User.php
```diff
@@ -289,4 +315,41 @@ public function scopeInactive($query)
...
+     * @param mixed $type
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopeAccessActive($query)
+    {
+        $now = new \DateTime();
+
+        return $query->whereHas('dues', function ($q) {
+            $q->paid()->accessCurrent();
+        })->orWhere('access_override_until', '>', date('Y-m-d'));
```
kberzinch: I don't know for sure if these date semantics are correct. Eventually, a job will be fired a few seconds after `access_override_until` and at that point it should see that the user's access flag is `false`.

It's okay to expire access slightly before intended (as the job won't fire until that time anyway) but it's not okay for the job to think a user has access when they don't (as you don't know that another job will fire later.)

kberzinch:
List of issues remaining on this PR:
 - [ ] CAS auth middleware triggers a `User` model save (and therefore a JEDI push) on _every_ request
   - This has been commented out for now
 - [ ] No jobs are triggered when a `User` is attached or detached from a `Team`
 - [ ] The `accessActive`/`accessInactive` scopes on the `User` model always return false if a user has an override that is expired
   - This is mitigated in the job
 - [ ] `SyncAccess` in Nova runs entirely in foreground (maybe fine?)

PR #581, opened 2019-03-02
Add Make a Wish Nova card

### LINKED ISSUE #576: Add card to Nova dashboard with Make a Wish link

JoshOldenburg:
Fix #576


JoshOldenburg:
image

JoshOldenburg:
One issue with this is that it's only on the dashboard, but I don't think we can edit Nova's footer.

PR #589, opened 2019-03-22
Fix #584

### LINKED ISSUE #584: Make recruiting visits in Nova not searchable
They pollute the search and aren't something searched for often

JoshOldenburg:


kberzinch:
Does this affect resource-level search as well as global?

JoshOldenburg:
Yes, it originally broke resource-level search, that's fixed now.

PR #802, opened 2019-08-29
Remove line about OrgSync group assignment from privacy policy

JoshOldenburg:


JoshOldenburg:
We don’t store it anymore

PR #804, opened 2019-09-01
Revert Square Connect version

JoshOldenburg:
The deprecation of a method we use is causing exceptions, so temporarily revert until after dues are due when it can be fixed properly.


kberzinch:
:x: PHP syntax errors were detected after running your Shift. Often these are simply differences between the PHP version of Shift and your project. Occasionally they are misplaced lines.

You can check the PHP syntax by running `php -l` on the following files:

- [ ] app/Http/Controllers/PaymentController.php

PR #814, opened 2019-09-10
Fix attendance kiosk admin

JoshOldenburg:
It was not using the /users/search endpoint, and the search endpoint was not including roles.


JoshOldenburg:
This also fixes the issue Dallas mentioned recently where when an unknown member swipes in, the next swipe gets discarded.

PR #837, opened 2019-10-01
Fix #834

### LINKED ISSUE #834: Add access active filter for users
I have this locally (it's just a copy of `UserActive` with some tweaks), but this is dependent on #833.

JoshOldenburg:


app/Nova/Filters/UserAccessActive.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Apply the filter to the given query.
+     *
+     * @param \Illuminate\Http\Request  $request
+     * @param \Illuminate\Database\Eloquent\Builder  $query
+     * @param string  $value
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function apply(Request $request, $query, $value): Builder
```
kberzinch: Can you type hint this?

JoshOldenburg: No, it seemed to not be valid since it doesn't match the parent signature.

kberzinch: Ah right that's a thing that happens.

PR #840, opened 2019-10-04
Send last attendance date to Jedi and sync necessary users daily

### LINKED ISSUE #824: Send last seen attendance swipe to JEDI
This may be used for SUMS lockout in a future release of JEDI, pending management approval and developer bandwidth.

JoshOldenburg:
Fixes #824.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -6,6 +6,7 @@ namespace App\Console;
 
 use App\Jobs\DailyDuesSummary;
 use App\Jobs\WeeklyAttendance;
+use App\Jobs\NoAttendanceJediPush;
 use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Foundation\Console\Kernel as ConsoleKernel;
 
@@ -31,6 +32,7 @@ class Kernel extends ConsoleKernel
 
         $schedule->job(new WeeklyAttendance())->weekly()->sundays()->at('11:00');
         $schedule->job(new DailyDuesSummary())->daily()->at('11:00');
+        $schedule->job(new NoAttendanceJediPush())->daily()->at('5:00');
     }
 
     /**
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -32,7 +32,7 @@ class Kernel extends ConsoleKernel
 
         $schedule->job(new WeeklyAttendance())->weekly()->sundays()->at('11:00');
         $schedule->job(new DailyDuesSummary())->daily()->at('11:00');
-        $schedule->job(new NoAttendanceJediPush())->daily()->at('5:00');
+        $schedule->job(new NoAttendanceJediPush())->daily()->at('10:00');
     }
 
     /**
--- a/app/Jobs/NoAttendanceJediPush.php
+++ b/app/Jobs/NoAttendanceJediPush.php
@@ -2,14 +2,16 @@
 
 declare(strict_types=1);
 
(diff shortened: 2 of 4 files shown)


app/Console/Kernel.php
```diff
@@ -31,6 +32,7 @@ protected function schedule(Schedule $schedule): void
 
         $schedule->job(new WeeklyAttendance())->weekly()->sundays()->at('11:00');
         $schedule->job(new DailyDuesSummary())->daily()->at('11:00');
+        $schedule->job(new NoAttendanceJediPush())->daily()->at('5:00');
```
kberzinch: I would rather send emails around 10am for better visibility.

app/Jobs/NoAttendanceJediPush.php
```diff
@@ -0,0 +1,36 @@
...
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle(): void
+    {
+        $users = User::active()->whereDoesntHave('attendance', static function (Builder $query) {
```
kberzinch: This should be `User::accessActive()`

app/Jobs/NoAttendanceJediPush.php
```diff
@@ -0,0 +1,36 @@
...
+            $query->where('attendable_type', Team::class)->where('created_at', '>', now()->startOfDay()->subDays(28));
```
kberzinch: I would like to avoid syncing a similar group of users to SUMS every day. We could either filter out any users that would have been sent in the previous day's job here, or I can bump up the time limit on the debounce cache to 48 hours.

app/Jobs/PushToJedi.php
```diff
@@ -76,8 +77,12 @@ public function handle(): void
         $send['model_class'] = $this->model_class;
         $send['model_id'] = $this->model_id;
         $send['model_event'] = $this->model_event;
-        $send['teams'] = [];
 
+        $lastAttendance = $this->user->attendance()->where('attendable_type', Team::class)
+            ->orderBy('created_at', 'desc')->first();
+        $send['last_attendance_time'] = $lastAttendance ? $lastAttendance->created_at : null;
```
kberzinch: Could you also send the attendance ID? I want to use that for dedup.

app/Jobs/NoAttendanceJediPush.php
JoshOldenburg: I'll filter by attendance created before 28 days ago and after 29 days ago, that should eliminate the duplication since it'll only match people for one day.

app/Console/Kernel.php
JoshOldenburg: Will try removing someone from the billing group while they're signed into a machine before pushing that back.

app/Jobs/PushToJedi.php
```diff
@@ -76,8 +77,13 @@ public function handle(): void
...
+        $send['last_attendance_id'] = $lastAttendance ? $lastAttendance->id : -1;
```
kberzinch:
```suggestion
        $send['last_attendance_id'] = $lastAttendance ? $lastAttendance->id : null;
```

app/Console/Kernel.php
JoshOldenburg: As discussed in Slack, removing someone from a billing group while they're signed in does not sign them out. They are unable to sign into the SUMS kiosk to sign out of the machine, but other people will be around who can.

PR #841, opened 2019-10-05
Add Nova action for sending NotificationTemplates to Users

kberzinch:


app/Nova/Actions/SendNotification.php
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return array<string,string>
+     */
+    public function handle(ActionFields $fields, Collection $models): array
+    {
+        foreach ($models as $model) {
+            Mail::to($model->gt_email)->send(new DatabaseMailable($fields->template, null));
+        }
+
+        return Action::message('Email'.(1 === count($models) ? '' : 's').' sent!');
```
JoshOldenburg: Use `Str::plural`

app/Nova/User.php
```diff
@@ -363,6 +363,13 @@ public function actions(Request $request): array
                 ->canRun(static function (Request $request, AU $user): bool {
                     return $request->user()->can('read-users');
                 }),
+            (new Actions\SendNotification())
+                ->canSee(static function (Request $request): bool {
+                    return $request->user()->hasRole('admin');
+                })
+                ->canRun(static function (Request $request, AU $user): bool {
+                    return $request->user()->hasRole('admin');
```
JoshOldenburg: Why not use the `send-notifications` permission?

PR #844, opened 2019-10-05
Make dues transaction actions visible on table rows

JoshOldenburg:


JoshOldenburg:
image
It's not perfect, since all the actions show up, but I think it's better because it saves a few clicks.

kberzinch:
This doesn't run the visibility logic either, so all the actions will show up for every transaction regardless of whether it's actually available.

JoshOldenburg:
We'll hold off on this until Nova runs the visibility logic then. (Hopefully they will implement that eventually.)

PR #909, opened 2019-10-19
Add several CAS session checks before authenticating users

### LINKED ISSUE #827: Add basic login checks from auth plugins
Check for domain, require duo, network check?

Consider reusing error pages package

kberzinch:
Closes #827


kberzinch:
This does not make any changes to the `CASCheck` middleware, not sure how/if we want to do that

kberzinch:
@RossIV can you test this with your various weird accounts

PR #941, opened 2019-10-31
Use GitHub Actions instead of CircleCI

JoshOldenburg:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    - run: sudo apt install libpng-dev libldap2-dev php7.2-zip php7.2-bcmath php7.2-gd php7.2-ldap
+    - name: Update Composer
+      run: sudo composer self-update
+    - name: Cache vendor directory
+      uses: actions/cache@preview
+      with:
+        path: vendor
+        key: ${{ runner.os }}-composer
+    - name: Install Composer dependencies
+      run: composer install --no-interaction --no-progress --no-suggest --no-dev
```
kberzinch: There are a couple more good flags to throw on here, look at the deployment script.

Also, you can pass `--ignore-platform-reqs` to get it to ignore extensions and some other things to save build time at L13. Since we don't actually execute much PHP code during build it should be fine.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+        COMPOSER_AUTH: ${{ secrets.COMPOSER_AUTH }}
+    - name: Check for env() being used outside config/
+      run: bash -c "cd $GIT_WORKSPACE && ! git grep \"env(\" -- '.' ':!/config/' ':!/.circleci/config.yml' ':!/composer.lock'"
+    - run: php artisan config:cache
+    - run: php artisan nova:publish --no-interaction
+    - run: php artisan horizon:assets --no-interaction
+    - name: Cache npm
+      uses: actions/cache@preview
+      with:
+        path: ~/.npm
```
kberzinch: What is in here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
kberzinch: Is there a better base image we could use?

In the long term we should roll our own Docker image imho but that doesn't have to be a this-PR problem.

.github/workflows/ci.yml
JoshOldenburg: That's npm's cache of downloads from the registry.

.github/workflows/ci.yml
JoshOldenburg: This is a VM template, not a Docker image, as I understand it. The options for that field are only Ubuntu 18.04, Ubuntu 16.04, Windows, and macOS.

kberzinch: Ah ok. From the docs it sounds like individual steps are docker containers though?

JoshOldenburg: At the very least we have the option to wrap it in a container.

PR #1061, opened 2020-01-18
Add ClickUp fields to users

### LINKED ISSUE #1021: Add OAuth for ClickUp accounts
https://clickup.com/api

Ship to JEDI for someone else (probably me) to deal with

### LINKED ISSUE #1052: Add SUMS as a linked account
This one is a bit weird since it's just a boolean flag on the account and a kick to JEDI Self Service, but it would help with discoverability

kberzinch:
This is dependent on upcoming changes to JEDI but should not cause problems before those are completed.

Fixes #1052 
Fixes #1021


resources/js/components/UserEditForm.vue
```diff
@@ -199,6 +240,10 @@ export default {
       .get(this.dataUrl)
       .then(response => {
         this.user = response.data.user;
+        this.clickUpEmailOptions = [
+          ...new Set([this.user.gt_email.toLowerCase(), this.user.uid.toLowerCase() + '@gatech.edu', this.user.gmail_address || this.user.gt_email.toLowerCase()])
```
kberzinch: This is wonky, open to suggestions

app/Nova/User.php
```diff
@@ -159,16 +159,27 @@ public function fields(Request $request): array
...
 
+                    Text::make('ClickUp', 'clickup_email')
+                        ->hideFromIndex()
+                        ->rules('nullable', 'max:255', 'email')
+                        ->creationRules('unique:users,clickup_email')
+                        ->updateRules('unique:users,clickup_email,{{resourceId}}'),
+
+                    Boolean::make('ClickUp Invite Pending', 'clickup_invite_pending')
+                        ->hideFromIndex()
+                        ->help('This flag is updated by JEDI in some cases but can be manually set if it is out of sync'
```
JoshOldenburg: Keep the wording consistent between this and the `exists_in_sums` help text.

app/Http/Controllers/ClickUpController.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    public function index(Request $request)
+    {
+        $user = $request->user();
+
+        if (! $user->is_access_active) {
+            return view(
+                'clickup',
+                [
+                    'message' => 'You have not paid dues, so you do not have access to ClickUp right now.',
```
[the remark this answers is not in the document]
kberzinch: That issue is poorly worded, I meant we should kick to JEDI if Apiary-side checks clear but there's an account already linked.

This is essentially the target state for that controller.

app/Nova/User.php
kberzinch: The difference is that the SUMS flag is guaranteed to be correct (at least as of last sync) and the ClickUp flag is not.

PR #1087, opened 2020-02-13
Fix PHPstan errors

kberzinch:


app/Http/Controllers/AttendanceController.php
```diff
@@ -119,11 +119,7 @@ public function search(SearchAttendanceRequest $request): JsonResponse
             ->start($request->input('start_date'))->end($request->input('end_date'))
             ->with($this->authorizeInclude(Attendance::class, $include))->get();
 
-        if ($att) {
-            return response()->json(['status' => 'success', 'attendance' => AttendanceResource::collection($att)]);
-        }
-
-        return response()->json(['status' => 'error', 'message' => 'Attendance not found.'], 404);
+        return response()->json(['status' => 'success', 'attendance' => AttendanceResource::collection($att)]);
```
JoshOldenburg: Why is this changing?

kberzinch: PHPStan was adamant that the call would always return a collection.

Issue #1136, opened 2020-03-13
GTED integration master plan


kberzinch:
**Code changes**
- [x] Add a `Major` model to represent a single major with the below attributes:
  - Display name (ex Computer Science), nullable
  - GTAD group name (ex `cs_cs`), not nullable
  - Whitepages OU (ex `CS`), nullable
- [x] Add a many-to-many relationship between `User`s and `Major`s
  - Relationship should have timestamps so that we can track when `Major`s were attached/detached from `User`s
- [x] Add a `create_reason` field to `User`s defaulting to `cas`(for initial data load) and then no default for normal operation
- [x] Add a `has_logged_in` field to `User`s defaulting to true (for initial data load) and then false for normal operation
- [x] Add a `primary_affiliation` field to `User`s
- [x] Add a `last_login_timestamp` field to `User`s defaulting to null
- [x] Represent the above information in Nova with options to edit
- [x] Add a `buzzapi` queue to Horizon
- [x] Add a job to create users given a GTID (#815) running in the `buzzapi` queue
- [x] Add a job to create users given a gatech.edu email address running in the `buzzapi` queue
- [x] Add a job to pull major data for a user given a GTID (for chaining with above) running in the `buzzapi` queue
- [x] Add a step during CAS login to update major information from attributes (probably want a `findOrCreate` method for `Major`s)
- [x] Queue user creation when recruiting visits are created
- [x] Queue user creation when attendance is created
- [x] Queue major update when users are created
- [x] Add API support (and documentation) for the `Major` model
- [x] Update resume book generation job to support cached `Major`s and make a `Major` selectable in the generation job

**One-off tasks**
- [x] Run user creation job for any attendance records with no user information
- [x] Run user creation job for any recruiting visit that can't already be associated to a user in the DB
- [x] Import dues payment information as far back as available
- [x] Run major pull job for all users

kberzinch:
All work has been completed, deprioritized, or split to separate issues.

PR #1137, opened 2020-03-15
Model changes for GTED integrations

kberzinch:
This pull request makes the database and model changes necessary to pull in affiliation and major information.

Only CAS information is actually used for data loading at this point; BuzzAPI will be added later.

See also #1136


app/Major.php
```diff
@@ -0,0 +1,58 @@
...
+
+declare(strict_types=1);
+
+namespace App;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+
+class Major extends Model
+{
```
JoshOldenburg: The soft deletes trait is missing

app/Major.php
```diff
@@ -0,0 +1,58 @@
...
+        'created_at',
+        'updated_at',
+    ];
+
+    /**
+     *  Get the Users that are members of this Major.
+     */
+    public function members(): BelongsToMany
+    {
+        return $this->belongsToMany(User::class)->whereNull('major_user.deleted_at')->withTimestamps();
```
JoshOldenburg: Is the `deleted_at` check required?

app/Nova/Major.php
```diff
@@ -0,0 +1,96 @@
...
+                ->sortable()
+                ->rules('required', 'max:255'),
+
+            BelongsToMany::make('Members', 'members', User::class),
+
+            new Panel(
+                'Metadata',
+                [
+                    DateTime::make('Created', 'created_at')
+                    ->onlyOnDetail(),
```
JoshOldenburg: Inconsistent indentation

app/Nova/Metrics/ActiveStudentsInMajor.php
```diff
@@ -0,0 +1,33 @@
...
+use Laravel\Nova\Metrics\ValueResult;
+
+class ActiveStudentsInMajor extends Value
+{
+    /**
+     * Calculate the value of the metric.
+     */
+    public function calculate(NovaRequest $request): ValueResult
+    {
+        $count = Major::where('id', $request->resourceId)->get()->first()->members()->active()->count();
```
JoshOldenburg: You don't need to fix it but `::find(id)` is an alias of `::where('id', id)->get->first()`. https://laravel.com/docs/5.8/eloquent#retrieving-single-models

app/User.php
```diff
@@ -213,6 +231,11 @@ public function rsvps(): HasMany
         return $this->hasMany(Rsvp::class);
     }
 
+    public function majors(): BelongsToMany
+    {
+        return $this->belongsToMany(Major::class)->whereNull('major_user.deleted_at')->withTimestamps();
```
JoshOldenburg: Same as above, is the `whereNull` check required?

app/User.php
```diff
@@ -374,4 +397,48 @@ public function accessOverrideBy(): BelongsTo
...
     }
+
+    /**
+     * Synchronizes major relationship with a given list of gtAccountEntitlements.
+     *
+     * @param array<string>  $accountEntitlements
+     */
+    public function syncMajorsFromAccountEntitlements(array $accountEntitlements): void
+    {
+        $current_major_ids = $this->majors()->get()->map(static function (Major $item): int {
```
JoshOldenburg: The `map` call could be simplified to `pluck('id')` but fixing it isn't necessary.

app/Major.php
kberzinch: I don't think it would be necessary to delete majors ever, soft or otherwise. The policy is also written to disallow it.

Do you still want to enable soft deletes?

app/Major.php
kberzinch: We're soft deleting relationships which laravel doesn't natively support, so yes.

app/Major.php
JoshOldenburg: No, it's fine without that. I thought the migration had added the column but I see that it hadn't. The `$guarded` array still contains `deleted_at` so I assumed it was meant to be there.

app/Major.php
```diff
@@ -0,0 +1,58 @@
...
+class Major extends Model
+{
+    /**
+     * The attributes that are not mass assignable.
+     *
+     * @var array<string>
+     */
+    protected $guarded = [
+        'id',
+        'deleted_at',
```
JoshOldenburg: `deleted_at` is unnecessary with a model that isn't deletable.

app/Major.php
kberzinch: Oops, yeah, that's a bad copy paste.

app/User.php
```diff
@@ -374,4 +397,46 @@ public function accessOverrideBy(): BelongsTo
     {
         return $this->belongsTo(self::class, 'access_override_by_id');
     }
+
+    /**
+     * Synchronizes major relationship with a given list of gtAccountEntitlements.
+     *
+     * @param array<string>  $accountEntitlements
+     */
+    public function syncMajorsFromAccountEntitlements(array $accountEntitlements): void
```
RossIV: Why is this in the user model? I feel like this would be better somewhere else - maybe a job?

database/migrations/2020_03_15_122454_add_metadata_fields_to_users.php
```diff
@@ -0,0 +1,50 @@
...
+    public function up(): void
+    {
+        DB::connection()->getDoctrineSchemaManager()->getDatabasePlatform()->registerDoctrineTypeMapping(
+            'enum',
+            'string'
+        );
+
+        Schema::table('users', static function (Blueprint $table): void {
+            $table->string('create_reason', 255)->default('cas_login');
+            $table->boolean('has_ever_logged_in')->default(true);
```
RossIV: Is this necessary if you have the `last_login` timestamp?

app/User.php
kberzinch: This will be used in multiple places including background jobs and during login.

database/migrations/2020_03_15_122454_add_metadata_fields_to_users.php
kberzinch: Tbh not really. The updated timestamp will always match this since it takes a database write to set that field.

JoshOldenburg: You can’t backfill the time stamp in the migration unless there’s an arbitrary date that means unknown but has logged in.

kberzinch: If the has logged in flag is set then the last updated timestamp is the last login timestamp, otherwise they haven't logged in ever.

PR #1150, opened 2020-03-21
Add class standings

kberzinch:


database/migrations/2020_03_21_150221_make_gtad_group_names_unique.php
```diff
@@ -0,0 +1,30 @@
...
+            $table->string('gtad_majorgroup_name')->unique()->change();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('geo', static function (Blueprint $table): void {
```
JoshOldenburg: Wrong table name? `geo`

kberzinch: Whoops, I copy-pasted that from the internet

PR #1153, opened 2020-03-21
Remote attendance

### LINKED ISSUE #1141: Add self-service attendance recording links
Short-lived links to say you were at a remote team meeting, similar to RSVPs with forced CAS auth

kberzinch:
Closes #1141

### DIFF
--- a/app/Nova/Actions/ResetRemoteAttendance.php
+++ b/app/Nova/Actions/ResetRemoteAttendance.php
@@ -29,10 +29,10 @@ class ResetRemoteAttendance extends Action
      */
     public function handle(ActionFields $fields, Collection $models): array
     {
-        $expiration = Carbon::now()->addHours(2);
+        $expiration = $fields['expiration_time'];
 
-        if (property_exists($fields, 'expiration_time')) {
-            $expiration = $fields->expiration_time;
+        if (null === $expiration) {
+            $expiration = Carbon::now()->addHours(1);
         }
 
         $team = $models->first();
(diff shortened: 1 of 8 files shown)


app/Nova/Actions/ResetRemoteAttendance.php
```diff
@@ -0,0 +1,64 @@
...
+     * Get the fields available on the action.
+     *
+     * @return array<\Laravel\Nova\Fields\Field>
+     */
+    public function fields(): array
+    {
+        if (Auth::user()->hasRole('admin')) {
+            return [
+                DateTime::make('Expiration Time', 'expiration_time')
+                    ->required(true)
```
JoshOldenburg: I'd suggest making this optional and saying in the help text that it defaults to 1 hour in the future.

app/Nova/Team.php
```diff
@@ -159,6 +161,39 @@ protected function controlFields(): array
...
+    {
+        return [
+            Text::make('Link', 'attendance_secret')
+                ->hideFromIndex()
+                ->resolveUsing(static function (?string $secret): ?string {
+                    return null === $secret ? null : config('app.url').'attendance/remote/'.$secret;
+                })
+                ->readonly(static function (Request $request): bool {
+                    return ! $request->user()->hasRole('admin');
+                })
```
JoshOldenburg: I don't think this should be editable at all. If it is going to be editable, there should be a separate Nova field for the secret itself with no custom resolve function so you don't end up with the attendance_secret having the `https://apiary-test.robojackets.org/attendance/remote/` prefix in it (which makes the link resolve as `https://apiary-test.robojackets.org/attendance/remote/https://apiary-test.robojackets.org/attendance/remote/randomcharactershere` if you don't change it).

kberzinch: In the interest of keeping admin superpowers 🙂 I'd like to keep it editable, though I will make a separate field as you suggest.

app/Nova/Actions/ResetRemoteAttendance.php
kberzinch: Heh, despite the red asterisk nova will let you run the action without specifying a timestamp and it does work as intended.

PR #1395, opened 2020-10-03
Add flag for hiding teams on kiosk and sort teams better

JoshOldenburg:
This is unimportant for the immediate future as the kiosk has been replaced by RoboJackets/people-counter. However, the sorting changes will make the teams page nicer.


kberzinch (approved):

routes/web.php
```diff
@@ -27,7 +27,7 @@
     Route::get('profile', 'UserController@showProfile');
 
     Route::prefix('dues')->group(static function (): void {
-        Route::get('/', 'DuesController@showDuesFlow')->name('payDues');
+        Route::get('/', 'DuesTransactionController@showDuesFlow')->name('payDues');
```
JoshOldenburg, line 30: This fixes an unrelated bug where the dues flow is broken.

PR #1512, opened 2020-12-22
Rebuild Square integration

kberzinch:
Summary of changes:
- Factor out Square-related endpoints to a separate controller
- Remove some routes that are no longer necessary
- Update to latest Square SDK
- Add webhook to receive payment events directly from Square

Closes #240 
Closes #448 
Closes #449 
Closes #480 
Closes #517 
Closes #518 
Closes #795
Closes #1164

Issue #830, opened 2019-09-17
PaymentsPerDay Nova card is invalid


JoshOldenburg:
It should use `updated_at`, not `created_at`, because `Payment`s are updated when the Square payment actually goes through. Additionally, it should ignore payments with a zero amount.

PR #1519, opened 2020-12-28
Historical dues and transaction data import

kberzinch:
Summary of changes:
- Adds `SquareCashTransaction` and `SquareTransaction` models and import commands for each
- Adds import command to map 2010-2017 dues spreadsheets to Apiary models
- Adds commands to match `SquareCashTransaction`s and `SquareTransaction`s to `Payment`s for backfilling all of the new fields that were added to `Payment`s in #1512 
- Adds additional error detection and better exception messages to `CreateOrUpdateUserFromBuzzAPI` job
- Add option to `CreateDuesPackages` Nova action to skip creating non-student package (intended for use with historical import only)


app/Models/Payment.php
```diff
@@ -134,4 +139,29 @@ public function getRelationshipPermissionMap(): array
...
+        $this->created_at = $transaction->transaction_timestamp;
+        $this->save();
+    }
+
+    public function updateFromSquareTransaction(SquareTransaction $transaction): void
+    {
+        $this->method = 'swipe';  // note that some POS transactions are actually keyed for some reason
+        $this->amount = $transaction->amount;
+        $this->processing_fee = $transaction->processing_fee;
+        $this->created_at = $transaction->transaction_timestamp;
```
kberzinch, line 158: It occured to me that we might want to add a `transaction_timestamp` to the `Payment` model in lieu of backdating these fields. That would also help with #830 somewhat.

app/Nova/Actions/CreateDuesPackages.php
```diff
@@ -142,6 +144,10 @@ public function handle(ActionFields $fields, Collection $models): array
      */
     public function fields(): array
     {
-        return [];
+        return [
+            Boolean::make('Non-Student Package', 'non_student')
+                ->help('Whether to create a package for non-students')
```
JoshOldenburg, line 149:
```suggestion
                ->help('Whether to create a package for non-students. This should generally be left checked.')
```

JoshOldenburg (changes requested):

app/Imports/HistoricalDues.php
```diff
@@ -0,0 +1,452 @@
...
+            if (null === $user) {
+                CreateOrUpdateUserFromBuzzAPI::dispatchNow('gtid', $gtid, self::USER_CREATION_REASON_STRING);
+                $user = User::where('gtid', $gtid)->firstOrFail();
+            }
+        }
+
+        foreach ($packages as $package) {
+            $guessShirtSize = self::guessShirtSize($row);
+            if (null !== $guessShirtSize) {
+                $user->shirt_size = $guessShirtSize;
```
kberzinch, line 114: Is it more correct to leave `polo_size` blank or copy the value from `shirt_size`?

app/Jobs/CreateOrUpdateUserFromBuzzAPI.php
```diff
@@ -136,6 +147,13 @@ public function handle(): void
         if ($user->is_service_account) {
             throw new Exception('BuzzAPI job attempted to create/update an account for an existing service account');
         }
+
+        foreach (self::EXPECTED_ATTRIBUTES as $attr) {
+            if (! property_exists($account, $attr)) {
+                throw new Exception('Selected account for '.$searchValue.' is missing expected attribute '.$attr);
+            }
+        }
```
kberzinch, line 155: This doesn't seem to work and I'm not sure why. The same code works in the import job.

app/Imports/HistoricalDues.php
JoshOldenburg: Leave it blank

app/Models/Payment.php
JoshOldenburg: Agreed that this would be good eventually but it doesn't need to happen for this PR.

JoshOldenburg (approved):

PR #1531, opened 2021-01-02
Membership agreements

### LINKED ISSUE #1265: Require corp liability waiver during dues flow
Lots of grooming to do on this but I think we should talk about it.

kberzinch:
Summary of changes:
- Adds `MembershipAgreementTemplate` and `Signature` models, and Nova resources for each
- Adds a `SignatureController` that handles generating paper copies of a `MembershipAgreementTemplate` and the full electronic signature flow
- Closes #1265

Known remaining work:
- [x] Drop `accept_safety_agreement` column from `users` table as that will be folded into the membership agreement
- [x] Prevent creating `DuesTransaction`s or accepting payments from users who do not have the latest `MembershipAgreementTemplate` signed
- [x] Send emails to the user's GT email and a corporate mailbox of some variety when a `Signature` transitions to `complete`
- [x] Feed agreement state to JEDI for turning off SUMS access
- [x] add username to hash
- [x] prevent access overrides
- [x] Add spot on print version to sign
- [x] Move UA/IP recording to render page instead of redirect

Intentionally out of scope:
- Removing the safety agreement from the dues flow


config/dompdf.php
```diff
@@ -0,0 +1,231 @@
...
+         * Dompdf's "chroot": Prevents dompdf from accessing system files or other
+         * files on the webserver.  All local files opened by dompdf must be in a
+         * subdirectory of this directory.  DO NOT set it to '/' since this could
+         * allow an attacker to use dompdf to read any files on the server.  This
+         * should be an absolute path.
+         * This is only checked on command line call by dompdf.php, but not by
+         * direct class use like:
+         * $dompdf = new DOMPDF();  $dompdf->load_html($htmldata); $dompdf->render(); $pdfdata = $dompdf->output();
+         */
+        'chroot' => realpath(base_path()),
```
JoshOldenburg, line 76: This chroot includes resumes and .env. Is this an issue?

app/Policies/SignaturePolicy.php
```diff
@@ -0,0 +1,70 @@
...
+class SignaturePolicy
+{
+    use HandlesAuthorization;
+
+    /**
+     * Determine whether the user can view any models.
+     */
+    public function viewAny(User $user): bool
+    {
+        return true;
```
JoshOldenburg, line 20: Do we really want these visible to everyone with Nova access?

app/Nova/Signature.php
```diff
@@ -0,0 +1,157 @@
...
+                        DateTime::make('Uploaded At', 'updated_at')
+                            ->resolveUsing(function (Carbon $str): ?Carbon {
+                                return null === $this->scanned_agreement ? null : $this->updated_at;
+                            })
+                            ->onlyOnDetail(),
+                    ]
+                ),
+            ] : []),
+
+            ...($this->electronic ? [
```
JoshOldenburg, line 114: You don't need the second ternary, you can put this panel in place of the previous ternary. I don't mind this as-is though.

app/Nova/MembershipAgreementTemplate.php
```diff
@@ -0,0 +1,87 @@
...
+        return [
+            ID::make('ID')
+                ->sortable(),
+
+            DateTime::make('Revision Date', 'updated_at')
+                ->sortable()
+                ->exceptOnForms(),
+
+            Markdown::make('Text')
+                ->help('Please only use h2\'s or smaller in this field.')
```
JoshOldenburg, line 81: This should specify the magic strings for the user's first and last name.

app/Http/Controllers/SignatureController.php
```diff
@@ -0,0 +1,285 @@
...
+
+        return redirect('/');
+    }
+
+    private function alreadySigned(Request $request, MembershipAgreementTemplate $template): bool
+    {
+        return null !== Signature::where('membership_agreement_template_id', $template->id)
+            ->where('user_id', $request->user()->id)
+            ->where('complete', true)
+            ->first();
```
JoshOldenburg, line 230: Instead of `first` and comparing to `null`, you can call `exists`.

JoshOldenburg (commented):
Looking good so far

app/Http/Controllers/SignatureController.php
JoshOldenburg: https://laravel.com/docs/8.x/queries#determining-if-records-exist

app/Policies/SignaturePolicy.php
kberzinch: 🤷🏻‍♂️ I don't think they're particularly sensitive. Do you want to restrict to PM+?

JoshOldenburg:
I think restricting it to `view-users` would be good enough.

app/Policies/SignaturePolicy.php
kberzinch: `read-users` includes everyone with Nova access

https://my.robojackets.org/nova/resources/permissions/2
https://my.robojackets.org/nova/resources/permissions/83

JoshOldenburg: It's fine as-is then. Never mind.

resources/views/agreement/alreadysigned.blade.php
```diff
@@ -0,0 +1,18 @@
...
+Membership Agreement | {{ config('app.name') }}
+@endsection
+
+@section('content')
+@component('layouts/title')
+  Membership Agreement
+@endcomponent
+
+<p>
+We already have a signed agreement on file for you for the most recent revision of the membership agreement.
```
ryanstrat, line 13: Might want to add a sentence explaining what the use should do next, even if it is "no action required"

I suspect the user might be somewhat confused if they ever end up here.

resources/views/agreement/render.blade.php
```diff
@@ -0,0 +1,38 @@
...
+
+<form method="POST" action="{{ route('agreement.redirect') }}">
+@csrf
+<div class="form-check">
+<input type="checkbox" class="form-check-input" id="electronicSignatureConsent" name="eSignConsent" required>
+<label class="form-check-label" for="electronicSignatureConsent">I consent to signing this document electronically. I understand that I can print a paper copy at <a href="{{ route('agreement.print') }}">this link</a> and submit it to a project manager or officer if I prefer not to sign electronically.</label>
+</div>
+<div class="form-check">
+<input type="checkbox" class="form-check-input" id="acknowledgement" name="acknowledgement" required>
+<label class="form-check-label" for="acknowledgement">I have read the above membership agreement and agree to it. I understand that clicking "Submit" below will send me to Georgia Tech Login, where I will need to log in again to electronically sign this document.</label>
```
ryanstrat, line 32: Prefer the language "I have read and agree to the above membership agreement."

JoshOldenburg (approved):

PR #1599, opened 2021-01-30
Improved validation and error handling for user edit forms

### LINKED ISSUE #746: Make sure people don't enter their own info in emergency contact

kberzinch:
This pull request fixes #746.

Summary of changes:
- Additional validation for `personal_email`, `phone`, `emergency_contact_name`, `emergency_contact_phone`, `clickup_email`, `autodesk_email`
- Gracefully(ish) handle 422 errors on the profile page and on the additional info page in the dues flow

### DIFF
--- a/app/Http/Requests/UpdateUserRequest.php
+++ b/app/Http/Requests/UpdateUserRequest.php
@@ -34,6 +34,7 @@ class UpdateUserRequest extends FormRequest
                 'max:255',
                 'nullable',
                 'unique:users,personal_email,'.$this->user()->id,
+                'email:rfc,strict,dns,spoof',
             ],
             'first_name' => [
                 'max:127',
@@ -48,13 +49,18 @@ class UpdateUserRequest extends FormRequest
                 'max:127',
             ],
             'phone' => [
-                'max:15',
+                'digits_between:10,15',
             ],
             'emergency_contact_name' => [
+                'nullable',
                 'max:255',
+                'required_with:emergency_contact_phone',
             ],
             'emergency_contact_phone' => [
-                'max:15',
+                'nullable',
+                'digits_between:10,15',
+                'different:phone',
+                'required_with:emergency_contact_name',
             ],
             'join_semester' => [
                 'max:6',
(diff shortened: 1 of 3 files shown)


app/Http/Requests/UpdateUserRequest.php
```diff
@@ -48,13 +49,18 @@ public function rules(): array
                 'max:127',
             ],
             'phone' => [
-                'max:15',
+                'digits_between:10,15',
             ],
             'emergency_contact_name' => [
+                'nullable',
```
JoshOldenburg, line 55: Do we really want to make this optional? Then we have to go back and make sure people filled it out before they go on a trip.

resources/js/components/UserEditForm.vue
```diff
@@ -315,6 +315,10 @@ export default {
           console.log('success');
         })
         .catch(response => {
+          if (response.response.status === 422) {
```
evan10s, line 318: To avoid issues with `null` values you probably want to write this as `response && response.response && response.response.status === 422` as I have learned from Typescript bugging me about not doing that many a time

Also, can you rename the parameter `response` to something else?

resources/js/components/UserEditForm.vue
```diff
@@ -315,6 +315,10 @@ export default {
...
+            Swal.fire('Invalid Data', response.response.data.errors[Object.keys(response.response.data.errors)[0]][0], 'error');
```
evan10s, line 319: Same here, this is a recipe for an `undefined` to cause all sorts of errors

JoshOldenburg (commented):
The code itself looks good to me. We should discuss further whether emergency contact should be optional.

evan10s (changes requested):
Just some JS nitpicking

app/Http/Requests/UpdateUserRequest.php
kberzinch: Most people that pay dues do not travel.

JoshOldenburg (approved):

evan10s (approved):

PR #1764, opened 2021-05-26
Add Passport to support OAuth2 API authentication

### LINKED ISSUE #1685: Update to Laravel Mix 6
I tried to do this but it was hairier than I expected.

evan10s:
To allow the future [Apiary mobile app](https://github.com/robojackets/apiary-mobile) to authenticate securely, I added OAuth2 using the [Laravel Passport](https://github.com/laravel/passport) package.

**Breaking changes:**
- ⚠️ The old API token functionality is no longer available.
  - Instead, you need to create a Personal Access Token (PAT).  PATs are valid for 1 year and the user associated with the PAT will receive emails prior to and at the time of the tokens expiration.  It's impossible to extend the lifetime of a PAT once created.
  - You can use PATs in the same way old API tokens were used to authenticate an API request.

**New stuff:**

- Added in Passport and updated routes/middleware so everything works
 - I had to add a new "login" route because of the way Passport's pre-defined middleware works.  (It's possible to change Passport's route middleware but that's more trouble than it's worth.)  `/login/cas` is behind the `auth.cas.force` middleware, so the route will get a user authenticated with CAS and then redirect them back to where they are going.
- With Passport added, most standard OAuth2 functionality is now supported.  We currently do not have any OAuth2 scopes defined.
  - Token expirations: Access: 1 day, Refresh: 1 month, PAT: 1 year
  - Expired and revoked tokens will be purged automatically twice per day.  In the event there are many more OAuth2 authentications occurring in the future, it may be desirable to increase the frequency of the cleanup job.
- Added new Nova functionality for:
  - Viewing or deleting/revoking a user's OAuth2 clients and token metadata (client secrets and actual token values are not visible after creation).
  - Creating a new OAuth2 client, either for the current user or (admins only) on behalf of another user
  - Creating a new personal access token for the current user or (admins only) another user
  - Delete a user's access/refresh tokens and/or PATs (current user or admins only)
- Added two new sections to the user profile page for showing authorized access tokens and personal access tokens:
  - This allows revoking access tokens (and associated refresh tokens) and personal access tokens
  - API endpoints exist from Passport to create clients and PATs but those aren't exposed in the UI currently
- Because OAuth2 Client IDs are UUIDs, I had to add a package that converts some ID columns on the `action_events` table to strings. From what I can tell, the action history in Nova works the same after the change.
- API routes are no longer accessible directly in web browsers due to complications with the Passport `auth:api` middleware.  You can still use an OAuth2 or personal access token, however.
- Adds Laravel Mix v6, closes #1685

**Screenshots:**

image

image

image

image

TODO:
- [x] Expiring PAT notification job
- [x] Fix CI
- [x] Write initial test/production server setup instructions
- [x] Figure out why editing Clients in Nova doesn't work
- [x] Add user-facing controls to revoke access/refresh tokens


app/Providers/AuthServiceProvider.php
```diff
@@ -76,5 +78,15 @@ class AuthServiceProvider extends ServiceProvider
...
+
+        if (! $this->app->routesAreCached()) {
+            Passport::routes();
+        }
+
+        Passport::useClientModel(PassportClient::class);
+        Passport::hashClientSecrets();
+        Passport::tokensExpireIn(now()->addDays(15));
+        Passport::refreshTokensExpireIn(now()->addDays(30));
+        Passport::personalAccessTokensExpireIn(now()->addMonths(6));
```
JoshOldenburg, line 90: I believe you said in Slack that you'd either extend this or add a job warning the user when tokens are about to expire.

JoshOldenburg (changes requested):

app/Http/Middleware/Authenticate.php
```diff
@@ -0,0 +1,22 @@
...
+     * Get the path the user should be redirected to when they are not authenticated.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return string
+     */
+    protected function redirectTo($request)
+    {
+        if (! $request->expectsJson()) {
+            return route('home');
+        }
```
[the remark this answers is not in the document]
evan10s: There are some instances where if you go through the OAuth process starting signed out, after you sign in, it takes you to the home page.  I'm fairly certain this middleware is causing that but I need to figure out how to make it redirect

But otherwise it not returning anything didn't seem to cause errors

config/passport.php
```diff
@@ -0,0 +1,66 @@
...
+    | Client UUIDs
+    |--------------------------------------------------------------------------
+    |
+    | By default, Passport uses auto-incrementing primary keys when assigning
+    | IDs to clients. However, if Passport is installed using the provided
+    | --uuids switch, this will be set to "true" and UUIDs will be used.
+    |
+    */
+
+    'client_uuids' => true,
```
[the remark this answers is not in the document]
evan10s: Was debating this, on one hand incrementing client IDs is much simpler, but on the other hand I had a feeling it'd be better to have more entropy with these values (and also prevent off-by-one errors).

database/migrations/2016_06_01_000001_create_oauth_auth_codes_table.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        $this->schema->create('oauth_auth_codes', function (Blueprint $table) {
+            $table->string('id', 100)->primary();
```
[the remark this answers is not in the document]
evan10s: Passport did this 😒, but I agree and I'll see if I can change it

config/passport.php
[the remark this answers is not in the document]
evan10s: Actually, on further inspection, this is just how Passport does its integer ID columns 🤦

database/migrations/2016_06_01_000001_create_oauth_auth_codes_table.php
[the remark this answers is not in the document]
evan10s: Actually, on further inspection, this is just how Passport does its integer ID columns 🤦 

The UUID columns are actually the `uuid` type though.  The column you're commenting on is an ID column in the auth codes table and those are just internal identifiers

app/Http/Middleware/Authenticate.php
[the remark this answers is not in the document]
evan10s: To clarify and expand what I sent in Slack, this is actually just overriding a function that's supposed to return a string.  It's part of an exception handler behind the scenes that's deciding where to send a user if they are logged out, basically

evan10s:
I'm moving this to a draft because I think Passport might not even be worth all the trouble and am probably going to back out and try again with Sanctum

app/Http/Controllers/AuthController.php
```diff
@@ -8,6 +8,19 @@
 
 class AuthController extends Controller
 {
+    /**
+     * Assuming this method is behind the `auth.cas.force` middleware, this function will ensure that the user is
```
JoshOldenburg, line 12: How does it enforce that with just a redirect to the previous page?

resources/views/mail/oauth2/pat_expiration.blade.php
```diff
@@ -0,0 +1,13 @@
+@component('mail::message')
+    @if($already_expired)
+        Your MyRoboJackets personal access token called {{ $token->name }} expired on {{ $token->expires_at }} and is no longer valid.
```
JoshOldenburg, line 3:
```suggestion
        Your MyRoboJackets personal access token called "{{ $token->name }}" expired on {{ $token->expires_at }} and is no longer valid.
```

resources/views/mail/oauth2/pat_expiration.blade.php
```diff
@@ -0,0 +1,13 @@
...
+    @else
+        Your MyRoboJackets personal access token called {{ $token->name }} will expire on {{ $token->expires_at }} and will not work after that time.
```
JoshOldenburg, line 5:
```suggestion
        Your MyRoboJackets personal access token called "{{ $token->name }}" will expire on {{ $token->expires_at }} and will not work after that time.
```

resources/views/mail/oauth2/pat_expiration.blade.php
```diff
@@ -0,0 +1,13 @@
...
+    @endif
+
+Personal access token expiration dates cannot be extended. If you're still using this token, you must create a new personal access token to continue accessing the MyRoboJackets API.  You can create a new personal access token by accessing your user page in Nova and running the Create Personal Access Token action.
```
JoshOldenburg, line 8:
```suggestion
Personal access token expiration dates cannot be extended. If you're still using this token, you must create a new personal access token to continue accessing the MyRoboJackets API.  You can create a new personal access token by accessing your user page in the MyRoboJackets admin interface and running the "Create Personal Access Token" action.
```

resources/views/oauth2clientcreated.blade.php
```diff
@@ -0,0 +1,45 @@
...
+            Your OAuth2 client was successfully created, and the client credentials are shown below.<br />
+            <small>Nova Actions can't show modals with information afterwards, so instead we had to send
+                you and your secrets to the Twilight Zone.</small>
+        </p>
+
+
+
+        @if(Session::has("client_plain_secret"))
+            <div class="alert alert-warning">
+                Save the client secret now—it will not be viewable after this!
```
JoshOldenburg, line 23:
```suggestion
                Save the client secret now&mdash;it will not be viewable after this!
```

resources/views/oauth2clientcreated.blade.php
```diff
@@ -0,0 +1,45 @@
...
+        <p>
+            @if(Session::get("client_confidential"))
+                <strong>Client Secret:</strong> {{ Session::get("client_plain_secret") ?? "Unavailable" }}
+            @else
+                <strong>Client Secret:</strong> This is a public (PKCE-enabled) client, so it doesn't have a client
+                secret.
+            @endif
+        </p>
+    @else
+        This page's content has expired.
```
JoshOldenburg, line 40:
```suggestion
        This page's content has expired. This might happen if you run the action quickly after loading the user's page due to a known technical limitation.
```

resources/views/personalaccesstokencreated.blade.php
```diff
@@ -0,0 +1,32 @@
...
+
+    @if(Session::has("pat_plain_token"))
+        <p>
+            Successfully created a new personal access token for {{ Session::get("pat_user_name") }}. <br />
+            <small>Nova Actions can't show modals with information afterwards, so instead we had to send
+                you and your secrets to the Twilight Zone.</small>
+        </p>
+
+        <div class="alert alert-warning">
+            Save this token now—it will not be viewable after this!
```
JoshOldenburg, line 20:
```suggestion
            Save this token now&mdash;it will not be viewable after this!
```

resources/views/personalaccesstokencreated.blade.php
```diff
@@ -0,0 +1,32 @@
...
+        </div>
+
+        <p class="text-break">
+            <strong>Personal Access Token:</strong>
+            <textarea readonly rows="11" onclick="this.select()" class="form-control text-break">{{ Session::get("pat_plain_token") ?? "Unavailable" }}</textarea></p>
+    @else
+        This page's content has expired.
```
JoshOldenburg, line 27:
```suggestion
        This page's content has expired. This might happen if you run the action quickly after loading the user's page due to a known technical limitation.
```

tests/Feature/PermissionsAndRolesTest.php
```diff
@@ -10,6 +10,16 @@
 
 class PermissionsAndRolesTest extends TestCase
 {
+    public function testPermissionsLoadedInDatabase(): void
+    {
+        $allPermissions = Permission::all();
+        // phpcs:disable
+        $this->assertGreaterThan(0, $allPermissions->count(), 'The permissions table appears to be '.
+            'empty, which means the database configuration for testing is not working. Are you running `composer run '.
```
JoshOldenburg, line 18:
```suggestion
            'empty, which means the database configuration for testing is not working, or a database dump was generated incorrectly (see https://github.com/RoboJackets/apiary/issues/1801). Are you running `composer run '.
```

JoshOldenburg (changes requested):
Looking pretty good so far. Most of my comments are for copy, not code. I'll give this a more thorough review soon but I don't expect to catch any real issues.

app/Http/Controllers/AuthController.php
evan10s: There's an expectation that you'll use this controller with appropriate middleware.  I added the `auth.cas.force` middleware to the route that uses this controller.

That being said, I've noticed it can cause a redirect loop if the user is unauthenticated so I need to add some extra logic to catch those situations.

evan10s: The redirect loop bug should be fixed now; it'll only redirect for the OAuth routes where this was needed

JoshOldenburg (approved):

app/Nova/OAuth2Client.php
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @return array<\Laravel\Nova\Fields\Field>
+     */
+    public function fields(Request $request): array
+    {
+        return [
+            ID::make(__('Client ID'), 'id')->sortable(),
+            Text::make('Name', 'name')->sortable(),
+            Boolean::make('Active', 'revoked')->trueValue(0)->falseValue(1),
+            Text::make('Redirect URL(s)', 'redirect')->onlyOnDetail(),
```
kberzinch, line 76: Is there a particular reason this is not editable?

app/Nova/OAuth2Client.php
```diff
@@ -0,0 +1,85 @@
...
+            Boolean::make('Public (PKCE-Enabled Client)', function (): bool {
+                return null !== $this->secret;
+            }),
+            Boolean::make('Password Client', 'password_client'),
```
kberzinch, line 80: Is this referring to the password grant type?

kberzinch (changes requested):

app/Nova/OAuth2Client.php
```diff
@@ -0,0 +1,85 @@
...
+    public static $displayInNavigation = false;
+
+    /**
+     * Get the fields displayed by the resource.
+     *
+     * @param \Illuminate\Http\Request  $request
+     *
+     * @return array<\Laravel\Nova\Fields\Field>
+     */
+    public function fields(Request $request): array
```
kberzinch, line 70: Add revoked field. Add policy to prevent deletes.

kberzinch (changes requested):
Additional notes per conversation in Slack

app/Providers/AuthServiceProvider.php
```diff
@@ -76,5 +77,14 @@ class AuthServiceProvider extends ServiceProvider
...
         $this->registerPolicies();
+
+        if (! $this->app->routesAreCached()) {
+            Passport::routes();
+        }
+
+        Passport::hashClientSecrets();
+        Passport::tokensExpireIn(now()->addDay());
+        Passport::refreshTokensExpireIn(now()->addMonth());
+        Passport::personalAccessTokensExpireIn(now()->addYear());
```
kberzinch, line 88: Add custom model to skip auth screen - https://laravel.com/docs/8.x/passport#approving-the-request

kberzinch (changes requested):

app/Nova/OAuth2Client.php
evan10s: Yes

kberzinch: I would rather pretend that doesn't exist. We could support it in theory but i have specifically put a lot of effort into not handling users' GT passwords.

app/Providers/AppServiceProvider.php
```diff
@@ -77,5 +78,6 @@ public function boot(): void
     public function register(): void
     {
         $this->app->alias('bugsnag.multi', \Psr\Log\LoggerInterface::class);
+        Passport::ignoreMigrations();
```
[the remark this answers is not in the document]
evan10s: They aren't, afaik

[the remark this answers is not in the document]
kberzinch: So why copy them into the repo...

app/Nova/OAuth2Client.php
evan10s: I added it for the sake of matching the model, but I'm pretty sure it wouldn't work anyways because the authentication end of it isn't implemented.

app/Nova/OAuth2Client.php
evan10s: I missed it, will fix

app/Nova/OAuth2Client.php
kberzinch: Right, my point is to remove something that doesn't work from the UI to simplify things.

app/Providers/AppServiceProvider.php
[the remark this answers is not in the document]
evan10s: at one point we did need it for adding a field to signify first-party clients. that's no longer a thing but the last time I tried to remove the migrations didn't go smoothly locally.  i'll see what I can do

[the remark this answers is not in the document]
kberzinch: I thought they should just get loaded out of the package. That's how the Nova and Horizon ones work, I think?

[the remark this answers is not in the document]
evan10s: If you use UUIDs for client IDs in Passport (which prevents easily guessing client IDs -- not that client IDs are intended to be secret -- things you can do by incrementing/decrementing an integer client ID by one, and client IDs revealing how many clients there are), it [automatically publishes the Passport migrations](https://github.com/laravel/passport/blob/10.x/src/Console/InstallCommand.php#L51) and replaces some text to use UUIDs as the primary key in the relevant spots.  Since it's publishing the migrations, I think it makes sense to ignore the Passport migrations.

Additionally, when Passport [makes changes that require updating the schema](https://github.com/laravel/passport/blob/10.x/UPGRADE.md#support-for-multiple-guards), they just tell you to add columns or whatever to the Passport tables, so either way you have to add migrations yourself 🙄

app/Models/Client.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Models;
+
+use Laravel\Passport\Client as PassportClient;
+
+class Client extends PassportClient
```
JoshOldenburg, line 9: Should this have a more specific name like `OAuth2Client` or `PassportClient`?

JoshOldenburg (commented):
One discussion point. The code changes are fine with me as long as Kristaps approves.

app/Models/Client.php
evan10s: good point, I agree. I'll rename it tomorrow night

kberzinch: Renamed to `OAuth2Client` to match other places

app/Providers/AppServiceProvider.php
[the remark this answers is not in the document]
kberzinch: Got it, makes sense.

kberzinch (approved):

PR #1853, opened 2021-07-19
Add Laravel Scout

### LINKED ISSUE #500: Integrate Meilisearch into Nova
Currently searching nova for a full name will return no results as all of the name columns are stored as parts. We did at one point have DB columns for full name but that has been moved to the PHP model and therefore isn't available for search in Nova.

kberzinch:
Fixes #500


app/Nova/User.php
```diff
@@ -517,4 +524,75 @@ public function actions(Request $request): array
...
+
+            return $firstYear === $lastYear ? 'Member '.$firstYear : 'Member '.$firstYear.'-'.$lastYear;
+        }
+
+        $major_names = $this->majors()->pluck('display_name')->toArray();
+
+        $class_standing_names = $this->classStanding()->pluck('name')->toArray();
+
+        if (count($major_names) > 0 && null !== $major_names[0] && count($class_standing_names) > 0) {
+            return $major_names[0].' | '.ucfirst($class_standing_names[0]);
```
evan10s, line 593: Remove these pls

composer.json
```diff
@@ -22,20 +22,24 @@
         "chelout/laravel-relationship-events": "1.4.0",
         "doctrine/dbal": "2.13.2",
         "guzzlehttp/guzzle": "7.3.0",
+        "http-interop/http-factory-guzzle": "^1.0",
         "laravel/framework": "8.50.0",
         "laravel/horizon": "5.7.9",
-        "laravel/passport": "*",
         "laravel/nova": "*",
+        "laravel/passport": "*",
```
evan10s, line 29: This isn't related to your changes, but we should specify a more exact version here (the latest Passport is 10.1.3)

evan10s (commented):
lgtm overall but I'd like to poke around at it in test a little bit more

composer.json
JoshOldenburg: Renovate will open a PR to do that after this is merged.

app/Nova/User.php
kberzinch: Not necessarily opposed, but why? Is there a different value you'd like for non-members?

composer.json
kberzinch: I think it doesn't touch `*` packages.

app/Nova/User.php
evan10s: Gah, GitHub mobile botched my comment location.  I meant the 2 `ray` lines above this

kberzinch: Fixed in c0ba56ec9a5b783245f8ee76f44098f786e01d2a

app/Models/Team.php
```diff
@@ -88,6 +89,7 @@ class Team extends Model
     use HasRelationshipObservables;
     use Notifiable;
     use SoftDeletes;
+    use Searchable;
```
evan10s, line 92: Does this model need to have ranking rules explicitly defined?

app/Models/TravelAssignment.php
```diff
@@ -70,6 +74,43 @@ class TravelAssignment extends Model
...
+     * The rules to use for ranking results in Meilisearch.
+     *
+     * @var array<string>
+     */
+    public $ranking_rules = [
+        'desc(user_revenue_total)',
+        'desc(user_attendance_count)',
+        'desc(user_signatures_count)',
+        'desc(user_recruiting_visits_count)',
+        'desc(user_gtid)',
```
evan10s, line 110: Not sure but sorting by user last name might be more intuitive?  I think the GTID sort would create the effect of a seemingly random ordering by name

app/Models/Team.php
kberzinch: I couldn't think of any that would be useful.

evan10s: image

It'd be nice if they appeared alphabetically

app/Models/TravelAssignment.php
kberzinch: Custom ranking rules need to be on numeric fields.

app/Models/Team.php
kberzinch: Add `attendance_count`

evan10s (approved):
lgtm

JoshOldenburg (approved):

PR #2125, opened 2021-10-30
Sentry improvements

### LINKED ISSUE #1947: Obfuscate access and refresh tokens before data is sent to Sentry
See https://sentry.io/organizations/robojackets/issues/2599291076/?project=5878752 for an example from test, but it's a security risk to have full tokens in Sentry

kberzinch:
- Scrub `refresh_token` field from request body if present (fixes #1947)
- Add user info if available


evan10s (dismissed):
Looks good.  But at some point we might want to update our privacy policy to mention that usernames are sent to Sentry with event reports (we need to do that for Google Play anyways).

evan10s (approved):

JoshOldenburg (approved):

PR #2150, opened 2021-11-13
Fix instructions for writing a check

kberzinch:


JoshOldenburg (approved):
This helps explain why so many checks are made out incorrectly.

PR #2262, opened 2022-01-18
Travel Authority Request tweaks

kberzinch:
- Add metrics for completion
- Use full name for traveler instead of preferred name
- Put primary contact instead of treasurer on travel index


evan10s (approved):
lgtm

PR #2307, opened 2022-02-05
Add accounting fields to TARs

### LINKED ISSUE #2295: Add accounting fields to TAR

kberzinch:
Closes #2295

### DIFF
--- a/app/Nova/Travel.php
+++ b/app/Nova/Travel.php
@@ -208,6 +208,22 @@ class Travel extends Resource
                             ' If this is not applicable, enter 0.'
                         )
                         ->hideFromIndex(),
+
+                    Text::make('Workday Project Number', 'tar_project_number')
+                        ->required()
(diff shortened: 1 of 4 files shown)


app/Nova/Travel.php
```diff
@@ -208,6 +208,22 @@ public function fields(Request $request): array
                             ' If this is not applicable, enter 0.'
                         )
                         ->hideFromIndex(),
+
+                    Text::make('Workday Project Number', 'tar_project_number')
```
JoshOldenburg, line 212: Would this make more sense as a dropdown?

JoshOldenburg (commented):
Looks good to me. One quick question

app/Nova/Travel.php
kberzinch: I feel like a dropdown would somewhat discourage getting the real answer from GT.

JoshOldenburg: Makes sense.

JoshOldenburg (approved):

PR #2359, opened 2022-02-26
Initial attempt at Dockerizing

kberzinch:
- Frontend built as first stage of a multi-stage build
- Primary image built on `debian:bullseye-slim`
  - Composer is copied directly to this image - not required at runtime but this seems like a simpler approach
  - Image will support both SQLite and MySQL databases
  - `git` is not installed which is causing some spurious logging messages since Sentry usually uses that to grab the version info - in production we will need to just provide that as an environment variable but I'll leave it for now
- Added SQLite database with testing user already having admin role
- Nova credentials are mounted as build-time secrets (requires Buildkit)

The Square SDK does not support PHP 8.1 as of the time of this writing, and the Debian/Ubuntu LTS/RHEL first-party repos don't have 8.0, so I'm sticking with 7.4 for now.


evan10s (commented):
Seems reasonable, but I don't feel like testing this locally atm

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      SESSION_DRIVER: file
+      QUEUE_DRIVER: sync
+      CAS_MASQUERADE: atesting3
+      CAS_MASQUERADE_gtGTID: 901234567
+      CAS_MASQUERADE_email_primary: developers@robojackets.org
+      CAS_MASQUERADE_givenName: Apiary
+      CAS_MASQUERADE_sn: Testing
+      CAS_MASQUERADE_gtAccountEntitlement: /gt/central/services/iam/two-factor/duo-user
+      CAS_MASQUERADE_authn_method: duo-two-factor
+      CAS_MASQUERADE_authnContextClass: mfa-duo
```
kberzinch, line 28: need to clean these up to match what prod cas is actually sending these days

Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+COPY --chown=www-data:www-data resources/ /app/resources/
+COPY --chown=www-data:www-data routes/ /app/routes/
+COPY --chown=www-data:www-data storage/ /app/storage/
+COPY --chown=www-data:www-data artisan composer.json composer.lock server.php /app/
+
+COPY --from=frontend --chown=www-data:www-data /app/public/ /app/public/
+
+WORKDIR /app/
+
+USER www-data
```
kberzinch, line 55: May need to add some additional configuration so that php-fpm master process runs as `www-data` instead of root. Should be fine for local use as-is though.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3.8"
+
+services:
+  apiary:
+    image: robojackets/apiary
+    network_mode: host
+    environment:
+      APP_NAME: Apiary Local
+      APP_ENV: local
+      APP_KEY: base64:JVvxt2ojHHrHVIHhTeCGkjWan3dhQQZwTKnEuUhxF0g=
```
RossIV, line 10: Why is `APP_KEY` hardcoded?

.nomad/apiary.nomad
```diff
@@ -0,0 +1,337 @@
...
+          timeout = "5s"
+        }
+
+        check_restart {
+          limit = 5
+          grace = "20s"
+        }
+
+        meta {
+          nginx-config = trimspace(trimsuffix(trimspace(regex_replace(regex_replace(regex_replace(regex_replace(regex_replace(regex_replace(regex_replace(regex_replace(trimspace(file("conf/nginx.conf")),"server\\s{\\s",""),"server_name\\s\\S+;",""),"root\\s\\S+;",""),"listen\\s.+;",""),"#.+\\n",""),";\\s+",";"),"{\\s+","{"),"\\s+"," ")),"}"))
```
RossIV, line 225: What is this function vomit?

config/healthcheck.php
```diff
@@ -13,16 +13,19 @@
...
         UKFast\HealthCheck\Checks\CacheHealthCheck::class,
         UKFast\HealthCheck\Checks\DatabaseHealthCheck::class,
         UKFast\HealthCheck\Checks\HttpHealthCheck::class,
         UKFast\HealthCheck\Checks\LogHealthCheck::class,
         UKFast\HealthCheck\Checks\MigrationUpToDateHealthCheck::class,
-        UKFast\HealthCheck\Checks\RedisHealthCheck::class,
-        UKFast\HealthCheck\Checks\SchedulerHealthCheck::class,
         UKFast\HealthCheck\Checks\StorageHealthCheck::class,
+        ...('google-play-review' === env('APP_ENV') ? [
+            OITNetworkServices\BuzzAPI\HealthChecks\CredentialsAreValid::class,
```
RossIV, line 23: Why add the BuzzAPI checks only for the reviewer instance?

.nomad/scripts/prestart.sh
```diff
@@ -0,0 +1,12 @@
...
+php artisan view:cache --no-interaction --verbose
+php artisan event:cache --no-interaction --verbose
+php artisan route:cache --no-interaction --verbose
+php artisan cache:clear --no-interaction --verbose
+php artisan migrate --no-interaction --force --verbose
+mkdir --parents /assets/${NOMAD_JOB_NAME}/
+cp --recursive --verbose public/* /assets/${NOMAD_JOB_NAME}/
+if [ ${PERSIST_RESUMES} = "false" ] && [ ${DB_CONNECTION} = "mysql" ]
+then
+    mysql --execute="update users set resume_date=null"
```
JoshOldenburg, line 11: I assume you've tested this, but in case you haven't, is a semicolon required here?

.nomad/scripts/set-job-name.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sed -i "/job/c\job \"$1\" {" apiary.nomad
```
JoshOldenburg, line 3: What does this do? A comment might be helpful here.

composer.json
```diff
@@ -137,9 +137,6 @@
         "classmap-authoritative": true,
         "preferred-install": "dist",
         "process-timeout": 20,
-        "platform": {
-            "php": "7.4.13"
```
JoshOldenburg, line 141: Wasn't this here for renovate?

database/schema/mysql-schema.dump
```diff
@@ -1295,5 +1295,3 @@ UNLOCK TABLES;
 /*!40111 SET SQL_NOTES=@OLD_SQL_NOTES */;
 
 -- Dump completed on 2021-04-26 23:12:04
-
-ALTER TABLE `users` ENCRYPTION='Y';
```
JoshOldenburg, line 1299: The associated instructions should be removed from the readme.

JoshOldenburg (changes requested):
The code looks fine to my untrained eye. Some parts could use a little documentation.

docker-compose.yml
kberzinch: This is just for local use, so it doesn't really matter. But I did learn a way to generate one on the fly and set it as an environment variable, so I'm doing that instead.

.nomad/apiary.nomad
kberzinch: This is essentially a lossless compression algorithm for Nginx configuration.

Meta tag values are limited to 512 characters, so this strips out whitespace, comments, and some lines that are populated by the nginx job itself (`server`, `server_name`, `listen`, `root`).

The point of all this is so Nginx configuration can be version-controlled and promoted between environments, like any other artifact.

config/healthcheck.php
kberzinch: BuzzAPI is flaky and I'm tired of getting spurious health check failures.

The Google Play review server, however, does not talk to real BuzzAPI, it hits a mock endpoint running on the same server, so I figured we may as well make sure that connectivity is working.

.nomad/conf/.env.tpl
```diff
@@ -0,0 +1,28 @@
+{{- range $key, $value := (key "apiary/shared" | parseJSON) -}}
+{{- $key | trimSpace -}}={{- $value | toJSON }}
+{{ end -}}
+CAS_HOSTNAME="sso.gatech.edu"
+CAS_REAL_HOSTS="sso.gatech.edu"
+CAS_SESSION_NAME="__Host-cas_session"
```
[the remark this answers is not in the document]
kberzinch: Assuming you're asking about the `__Host-` prefix - it indicates to the browser that this cookie should meet certain criteria and reject it if not. [[mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes)]

.nomad/conf/.env.tpl
```diff
@@ -0,0 +1,28 @@
...
+MAILGUN_DOMAIN="my.robojackets.org"
+MAILGUN_SECRET="{{- key "mailgun/secret" | trimSpace -}}"
+SESSION_SECURE_COOKIE="true"
+SESSION_COOKIE="__Host-apiary_session"
+{{ range $key, $value := (key (printf "apiary/%s" (slice (env "NOMAD_JOB_NAME") 7)) | parseJSON) -}}
+{{- $key | trimSpace -}}={{- $value | toJSON }}
+{{ end -}}
+APP_ENV="{{ slice (env "NOMAD_JOB_NAME") 7 }}"
+APP_URL="https://{{- with (key "nginx/hostnames" | parseJSON) -}}{{- index . (env "NOMAD_JOB_NAME") -}}{{- end -}}"
+HOME="/secrets/"
```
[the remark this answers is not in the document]
kberzinch: I'm writing a MySQL config file to `/secrets/.my.cnf` with the app's database credentials, so I can run `mysql` in a Nomad web shell and get a MySQL prompt without any further configuration.

I'm setting `/secrets/` as the home directory because that's the easiest place to mount the files that MySQL will search. [[mysql docs](https://dev.mysql.com/doc/refman/8.0/en/option-files.html#:~:text=Table%C2%A04.2%C2%A0Option%20Files%20Read%20on%20Unix%20and%20Unix%2DLike%20Systems)]

.nomad/scripts/prestart.sh
kberzinch: This has indeed been tested and it works without a semicolon.

.nomad/scripts/set-job-name.sh
kberzinch: It is not currently possible to set the job ID dynamically: https://github.com/hashicorp/nomad/issues/9522

This script is a quick way to replace the job ID on disk before submitting the job. It is not used for anything yet, but will be integrated into the CD jobs at some point.

composer.json
kberzinch: Yeah, I don't think it did anything though. PHP 8.1 is next on the agenda regardless.

database/schema/mysql-schema.dump
kberzinch: Updated.

PR #2466, opened 2022-05-08
Pre-compress static assets and add CSP header

### LINKED ISSUE #2465: Asset improvements (pre-compress, crossorigin anonymous, CSP)

kberzinch:
- Compress JS, CSS, SVG, and PNG files inside the Docker image and serve them with `gzip_static`
- Add `Content-Security-Policy` header

Closes #2465


kberzinch:
This is running in test, GitHub just doesn't know about it. (Maybe I should add a job in Concourse...)

JoshOldenburg (approved):

Dockerfile
```diff
@@ -61,4 +67,15 @@ RUN --mount=type=secret,id=composer_auth,dst=/app/auth.json,uid=33,gid=33,requir
     composer install --no-interaction --no-progress --no-dev --optimize-autoloader --classmap-authoritative --no-cache && \
     php artisan nova:publish && \
     php artisan horizon:publish && \
-    sed -i '/HTTPS_ONLY_COOKIES/c\true,' /app/vendor/subfission/cas/src/Subfission/Cas/CasManager.php
+    sed -i '/HTTPS_ONLY_COOKIES/c\true,' /app/vendor/subfission/cas/src/Subfission/Cas/CasManager.php && \
```
RossIV, line 70: Why are you modifying vendor package files?

Dockerfile
```diff
@@ -2,23 +2,38 @@
 
 ARG base_image="debian:bullseye-slim"
 
+FROM scratch as frontend-source
+
+# artisan is not strictly required for JS builds but it triggers some behavior inside Laravel Mix
+# https://github.com/laravel-mix/laravel-mix/issues/1326#issuecomment-363975710
+COPY --link package.json package-lock.json webpack.mix.js artisan /app/
```
evan10s, line 9: purely out of curiosity since I can't seem to find it in the docs, what does the `--link` here do?

Dockerfile
kberzinch: That replaces this parameter with a hardcoded `true`: https://github.com/subfission/cas/blob/master/src/Subfission/Cas/CasManager.php#L46

This field cannot be set at runtime because `env` does not work when config is cached.

That should definitely be fixed upstream but I didn't want to deal with it at the time I wrote this.

evan10s (approved):

Dockerfile
kberzinch: It is hard to find (it's [here](https://hub.docker.com/r/docker/dockerfile#:~:text=Linked%20copies%20COPY%20%2D%2Dlink%2C%20ADD%20%2D%2Dlink)), but basically it improves caching across builds.

RossIV (approved):

Dockerfile
kberzinch: Opened #2475 to track

PR #2494, opened 2022-05-15
Performance improvements

### LINKED ISSUE #2458: Improve Nova performance
https://github.com/laravel/nova-issues/issues/2663

### LINKED ISSUE #2468: Look at optimizing dashboard load
From Sentry data it looks like a lot of queries are being run more than once. Cache results to reduce load time.

### LINKED ISSUE #2479: Add link headers to get browsers fetching assets as early as possible

kberzinch:
Fixes #2468
Fixes #2458

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -802,7 +802,7 @@ class User extends Authenticatable
     {
         $uid = $this->uid;
 
-        return Cache::remember('home_department_'.$uid, now(), static function () use ($uid): ?string {
+        return Cache::remember('home_department_'.$uid, now()->addDay(), static function () use ($uid): ?string {
             $parentSpan = SentrySdk::getCurrentHub()->getSpan();
 
             if (null !== $parentSpan) {
--- a/app/Providers/NovaServiceProvider.php
+++ b/app/Providers/NovaServiceProvider.php
@@ -26,7 +26,6 @@ use Carbon\Carbon;
 use Illuminate\Database\Query\JoinClause;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Gate;
-use Illuminate\Support\Str;
 use Laravel\Nova\Events\ServingNova;
 use Laravel\Nova\Nova;
 use Laravel\Nova\NovaApplicationServiceProvider;
--- a/app/Providers/NovaServiceProvider.php
+++ b/app/Providers/NovaServiceProvider.php
@@ -73,7 +73,8 @@ class NovaServiceProvider extends NovaApplicationServiceProvider
      */
     protected function cards(): array
     {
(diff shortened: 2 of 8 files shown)


app/Models/User.php
```diff
@@ -802,7 +802,7 @@ public function getHomeDepartmentAttribute(): ?string
     {
         $uid = $this->uid;
 
-        return Cache::remember('home_department_'.$uid, now(), static function () use ($uid): ?string {
+        return Cache::remember('home_department_'.$uid, now()->addDay(), static function () use ($uid): ?string {
```
JoshOldenburg, line 805: Nice

app/Providers/NovaServiceProvider.php
```diff
@@ -72,6 +73,11 @@ protected function gate(): void
      */
     protected function cards(): array
     {
+        if ('/nova-api/dashboards/main' !== $_SERVER['REQUEST_URI'] &&
+            ! Str::startsWith($_SERVER['REQUEST_URI'], '/nova-api/metrics/')) {
```
JoshOldenburg, line 77: There's not a Laravel way to get that variable?

JoshOldenburg (approved):

app/Providers/NovaServiceProvider.php
kberzinch: Maybe, let me try something

app/Models/User.php
evan10s, line 805: It seems like the second parameter of this function, `now()...`, is the duration to store the item in the cache if the item isn't present.  Unless you can give it a date, but that's not what it looks like based on how [the docs](https://laravel.com/docs/8.x/cache#retrieve-store) describe this function.  How I'm reading it, it seems like you'd want `24*60*60` for the seconds value.

config/htmlmin.php
```diff
@@ -0,0 +1,72 @@
...
+    |--------------------------------------------------------------------------
+    | Ignore Blade Files
+    |--------------------------------------------------------------------------
+    |
+    | Here you can specify paths, which you don't want to minify.
+    |
+    */
+
+    'ignore' => [
+        'resources/views/emails',
```
evan10s, line 62: Why can't we minify these things?

kberzinch: That was just the default config. I don't think it matters much either way.

app/Models/User.php
kberzinch: The [source](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Facades/Cache.php#L22) indicates that a `\DateTimeInterface` can be provided in that field.

evan10s (approved):

JoshOldenburg (approved):

PR #2496, opened 2022-05-16
Nova performance improvements

kberzinch:
- phpCAS sessions in Redis
- Add additional FPM children in prod
- Don't add unnecessary whitespace in HTML
- Cache Nova permission for 24 hours
- Lazy-load resources in Nova provider

### DIFF
--- a/app/Providers/NovaServiceProvider.php
+++ b/app/Providers/NovaServiceProvider.php
@@ -62,7 +62,9 @@ class NovaServiceProvider extends NovaApplicationServiceProvider
     protected function gate(): void
     {
         Gate::define('viewNova', static function (User $user): bool {
-            return $user->can('access-nova');
+            return Cache::remember('can_access_nova_'.$uid, now()->addDay(), static function () use ($user): ?string {
+                return $user->can('access-nova');
+            });
         });
     }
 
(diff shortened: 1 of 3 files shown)


app/Providers/NovaServiceProvider.php
```diff
@@ -61,7 +63,13 @@ protected function routes(): void
     protected function gate(): void
     {
         Gate::define('viewNova', static function (User $user): bool {
-            return $user->can('access-nova');
+            return Cache::remember(
```
evan10s, line 66: Will Nova access changes still take effect immediately?

evan10s (dismissed):

app/Providers/NovaServiceProvider.php
kberzinch: No, this would only get flushed during deployments.

evan10s: I guess it's just something to keep in mind when giving new people access to Nova for the first time. But it sounds like that should be a case where the cache key for a user should be flushed, if that's a thing.

kberzinch: I don't think there's a good way to do that right now.

evan10s: It would take a bit of effort but a listener for role assignment for the common roles that grant Nova access would probably do it. I would at least make an issue for that.

kberzinch: Opened #2497

JoshOldenburg (approved):

evan10s (approved):

PR #2536, opened 2022-05-28
Upgrade to Laravel Nova 4

### LINKED ISSUE #2464: Upgrade Nova to latest version

kberzinch:
Closes #2464


app/Nova/Signature.php
```diff
@@ -147,14 +144,6 @@ public function fields(Request $request): array
             ]),
 
             self::metadataPanel(),
-
-            Hidden::make('Complete')
```
JoshOldenburg, line 151: Is there an issue open to replace this?

JoshOldenburg (dismissed):

app/Nova/Signature.php
kberzinch: No, but I don't think we need one. Almost no one uses this workflow and I want to completely replace it with DocuSign this summer anyway.

evan10s (dismissed):
Seems reasonable

PR #2552, opened 2022-05-30
Show dues reminders in Nova

kberzinch:


app/Jobs/CreateDuesPaymentDueNotificationInNova.php
```diff
@@ -0,0 +1,66 @@
...
+    public function handle()
+    {
+        if ($this->user->dues()->pending()->count() > 0 &&
+            $this->user->hasPermissionTo('access-nova') &&
+            0 === Notification::where('notifiable_type', '=', User::class)
+                ->where('notifiable_id', '=', $this->user->id)
+                ->where('type', DuesPaymentDue::class)
+                ->where('created_at', '>', now()->subMonths(3))
+                ->count()
+        ) {
```
evan10s, line 52: Might be a decent candidate for a query scope on the User class?

app/Notifications/Nova/DuesAreLive.php
```diff
@@ -0,0 +1,64 @@
...
+     * @var string|null
+     */
+    public $message = 'Dues are live! Take a moment to pay online now.';
+
+    /**
+     * The text used for the call-to-action button label.
+     *
+     * @var string
+     */
+    public $actionText = 'Get started';
```
evan10s, line 37: I'd prefer action-oriented text on this button as well, but this `Dues are live` notification type seems like overlaps a lot with the `Dues payment due` notification.  I almost feel like we should only show one or the other.

kberzinch: `DuesAreLive` means you need to create a `DuesTransaction`, `DuesPaymentDue` means you need to complete a `Payment`. The prune job will delete a `DuesAreLive` notification once a `DuesTransaction` is created.

app/Jobs/CreateDuesPaymentDueNotificationInNova.php
kberzinch: I don't think a query scope is quite the right refactor, but I will make it an Eloquent relationship.

app/Notifications/Nova/DuesAreLive.php
evan10s: That all makes sense now, might be good to leave a quick note with the intended distinction in a comment on one/both of these classes for posterity

kberzinch: Added comments to both classes.

evan10s (approved):

PR #2556, opened 2022-05-31
Add self-service overrides

### LINKED ISSUE #1351: Support self-service access overrides
Per Slack discussion, show a dashboard card for anyone who has never paid dues or had an override in the past after they attend a meeting. The card has a button to grant the user an override until the next future access end date so long as that date is within the next two months, otherwise it can’t be given out.

evan10s:
Closes #1351.  I documented all the nuances below so they can eventually be transferred to the wiki if we so desire.

Self-service overrides allow RoboJackets newcomers to access online RoboJackets systems before they pay dues.  To prevent abuse, several criteria must be met before a user becomes eligible for a self-service override.  Under normal circumstances, it should be very easy for a well-intentioned RoboJackets new member to receive a self-service override.

Self-service override criteria are divided into 2 categories: conditions and tasks.  A condition is a system- or user-level property that cannot be rectified by the user alone (see list below).  However, tasks can be completed by users.

Required system/user conditions:
- User access must not be active
- User must have no prior dues payments
- User must have no previous access override
- A future dues package must exist

Required tasks:
- Sign the membership agreement
- Attend a team meeting

A user's eligibility for a self-service override is always visible on their details page in Nova in the System Access panel.  Additionally, when a user is eligible or can complete tasks to become eligible for a self-service override, a card will appear on their dashboard with a link to activate the override (otherwise, the card won't appear).

Self-service overrides can last between 7 and 60 days.  The longest duration will occur if the next (earliest) future dues package access end date is 60 or more days in the future.  The shortest duration will occur if the next future access end date is 7 or less days away.

image

self_service_override_demo


app/Models/SelfServiceAccessOverrideEligibility.php
```diff
@@ -0,0 +1,203 @@
+<?php
+
+namespace App\Models;
```
JoshOldenburg, line 3: I'm not sure how I feel about this living in Models when it's not in the database. I don't have a better suggestion though.

app/Http/Controllers/DashboardController.php
```diff
@@ -82,7 +82,7 @@ public function index(Request $request)
             $firstPayment = null;
         }
 
-        $hasOverride = ! $user->is_active && $user->access_override_until && $user->access_override_until > now();
+        $hasOverride = $user->has_active_override;
```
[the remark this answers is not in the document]
kberzinch: Yeah, there's a check for `is_active` in the getter.

app/Models/SelfServiceAccessOverrideEligibility.php
kberzinch: Same on both counts. I think it's fine.

app/Models/User.php
```diff
@@ -951,4 +954,81 @@ public function toSearchableArray(): array
...
+        $overrideEndDate = null;
+
+        // conditions
+        $accessNotActive = ! $this->is_access_active;
+        $noExistingOverride = null === $this->access_override_until;
+        $hasNotPaidDues = ! $this->paidDues()->exists();
+        $eligibleDuesPkgExists = null !== $nextAccessEndDuesPkg;
+
+        // tasks
+        $attendedTeamMeeting = $this->attendance()->whereAttendableType('team')->exists();
```
kberzinch, line 991: Should this exclude training teams?

evan10s: We usually only give out overrides in the context of the actual teams so I suppose yes.  The trainings outside of Apiary aren't really in the context of a team, so I think it's fine to add the restriction and note in documentation but keep the language as is otherwise.

kberzinch: I guess also exclude the various versions of core, to be thorough

evan10s: Sounds like we should add a property to teams rather than base this off team name, imo.

kberzinch: sgtm

resources/views/welcome.blade.php
```diff
@@ -167,38 +167,37 @@
...
+                            </p>
+                            <a href="{{ $travelAuthorityRequestUrl }}">Submit Now</a> or <a href="{{ route('travel.index') }}">View Travel</a>
+                        </div>
                     </div>
                 </div>
-            </div>
-        @endif
+            @endif
         @endif
+        <self-service-access-override />
```
kberzinch, line 201: Is there a particular reason this is rendered client side?

routes/api.php
```diff
@@ -39,6 +39,7 @@ static function (): void {
         Route::resource('users', UserController::class)->except('create', 'edit');
         Route::post('users/{id}/resume', [ResumeController::class, 'store']);
         Route::get('user', [UserController::class, 'showSelf']);
+        Route::post('user/override/auto', [UserController::class, 'applySelfOverride']);
```
kberzinch, line 42: I don't think `auto` makes sense here, maybe `self` instead.
```suggestion
        Route::post('user/override/self', [UserController::class, 'applySelfOverride']);
```

tests/Feature/SelfServiceOverrideTest.php
```diff
@@ -0,0 +1,297 @@
...
+class SelfServiceOverrideTest extends TestCase
+{
+    /**
+     * Shortcut to create a dummy dues package.
+     *
+     * @param  CarbonImmutable|null  $base_date  Date around which the dues package's validity periods will be defined
+     * @param  float  $cost
+     * @return DuesPackage
+     */
+    public function createDuesPackage(?CarbonImmutable $base_date, float $cost = 10): DuesPackage
```
kberzinch, line 31: There should be a factory for this already

kberzinch (changes requested):

resources/views/welcome.blade.php
evan10s: It enables the swal dialog to show info after requesting the override.  It is theoretically possible for this to be server-side, but I think the UX is smoother this way.  I could probably clean up some of the dashboard logic related to overrides since I didn't end up using that.

app/Models/User.php
evan10s: On further consideration, I don't think this restriction is worth implementing.  I don't like the idea of adding such a specific property and don't think it's necessarily problematic that people will become eligible for a self-service override from attending a training.

kberzinch: Fine with me

resources/js/components/dues/SelfServiceAccessOverride.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <div class="col-sm-6 com-md-3 col-lg-4"
+       v-if="!loading && eligibility && (eligibility.eligible || eligibility.user_rectifiable)">
+    <div class="card">
+      <div class="card-body">
+        <h4 class="card-title">
+          New Member Activation
```
kberzinch, line 7:
```suggestion
          New Member Access
```
care amount=6

resources/js/components/dues/SelfServiceAccessOverride.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <div class="col-sm-6 com-md-3 col-lg-4"
+       v-if="!loading && eligibility && (eligibility.eligible || eligibility.user_rectifiable)">
```
kberzinch, line 3: Idk how I feel about the card showing up for membership agreements. There's already a (hopefully very clear) card for that.

Care amount=4

kberzinch (approved):

resources/js/components/dues/SelfServiceAccessOverride.vue
evan10s: The reason to show both is because while the self-service override doesn't appear, it's not apparent what has to be done to receive one

kberzinch: I feel like "Start Here" is pretty apparent, but I'm not going to push on this.

kberzinch (approved):

JoshOldenburg (approved):

PR #2578, opened 2022-06-04
Add user impersonation

### LINKED ISSUE #2073: Implement admin masquerade-on-demand option for debugging
Allowing admins to act like another user (like Canvas) could speed up debugging significantly. Enter the mode with a Nova action on a user. There should be a banner (or similar) to make it obvious when you're in that mode. This needs some thought to avoid introducing weird/inconsistent data to the DB.

https://github.com/404labfr/laravel-impersonate

### LINKED ISSUE #2540: User menu in Nova no longer includes profile and logout links

kberzinch:
Closes #2073
Closes #2540 

This does not include any specific indication that you are impersonating another user, but given that (a) it is difficult to trigger accidentally and (b) the current user's name is on (almost) every page in the navbar, I don't know if that is strictly necessary.

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace App\Http\Controllers;
 
+use App\Models\User;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+use Laravel\Nova\Contracts\ImpersonatesUsers;
 
 class AuthController extends Controller
 {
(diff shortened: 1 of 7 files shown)


JoshOldenburg (changes requested):
I do think it should have some indication that you're impersonating a user. It could be as simple as turning the name in the nav bar red.

kberzinch:
Sigh, makin' me do frontend work

JoshOldenburg (approved):

JoshOldenburg (approved):

PR #2582, opened 2022-06-04
New payment receipt email

kberzinch:
Replaces the existing payment receipt email with a plain-text version.


database/factories/TravelFactory.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Define the model's default state.
+     *
+     * @return array<string,string|int|\DateTime>
+     */
+    public function definition()
+    {
+        return [
+            'name' => $this->faker->word(),
+            'primary_contact_user_id' => User::all()->random()->id,
```
evan10s, line 26: I'd prefer something that creates a user if one doesn't exist. Or does this do that?

database/factories/TravelFactory.php
```diff
@@ -0,0 +1,34 @@
...
+            'destination' => $this->faker->word(),
+            'departure_date' => $this->faker->dateTimeBetween('now', '1 year'),
+            'return_date' => $this->faker->dateTimeBetween('now', '1 year'),
```
evan10s, line 29: There's no guarantee here that return date is after departure date.  Maybe adjust the time ranges for departure/return date so they can't overlap the wrong way.

resources/views/mail/paymentreceipt.blade.php
```diff
@@ -0,0 +1,11 @@
+Hi {{ $payment->payable->user->preferred_first_name }},
+
+@if($payment->method === "square")
+We received your online payment of ${{ number_format($payment->amount, 2) }} for {{ $payable_name }}.
+
+You can view more details about your payment at {{ $payment->receipt_url }}
```
evan10s, line 6:
```suggestion
You can view more details about your payment at {{ $payment->receipt_url }}.
```

JoshOldenburg (approved):

database/factories/TravelFactory.php
kberzinch: This picks a random user. That's what the `DuesTransactionFactory` was already doing so I copied that.

I think this is reasonable default behavior; if something more specific is required for a test then the attribute can be overriden - https://laravel.com/docs/9.x/database-testing#overriding-attributes

resources/views/mail/paymentreceipt.blade.php
kberzinch: Will need to test how that renders in real clients, but ok

evan10s (approved):

JoshOldenburg (approved):

PR #2588, opened 2022-06-06
Tweak welcome language on home page

evan10s:
Adds logic for when to show "Welcome, George!" vs. "Welcome back, George!" ("Welcome back" will appear if the user has ever signed a membership agreement.)

Moves the "Welcome to RoboJackets" language from the New Member Access (self-service override) card to the Start Here (membership agreement) card.

Removes self-service override calculation from the dashboard template since it's unused.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -15,7 +15,7 @@
             <div class="card">
                 <div class="card-body">
                     <h4 class="card-title">
-                        Welcome {{ $signedAnyAgreement ? "back" : "" }}, {{ $preferredName }}!
+                        Welcome{{ $signedAnyAgreement ? " back" : "" }}, {{ $preferredName }}!
                     </h4>
                     <p class="card-text">
                         @if($isNew)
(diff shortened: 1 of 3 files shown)


resources/views/welcome.blade.php
```diff
@@ -15,7 +15,7 @@
             <div class="card">
                 <div class="card-body">
                     <h4 class="card-title">
-                        Welcome back, {{ $preferredName }}!
+                        Welcome {{ $signedAnyAgreement ? "back" : "" }}, {{ $preferredName }}!
```
kberzinch, line 18:
```suggestion
                        Welcome{{ $signedAnyAgreement ? " back" : "" }}, {{ $preferredName }}!
```

kberzinch (changes requested):

resources/views/welcome.blade.php
evan10s: Good catch

JoshOldenburg (approved):

kberzinch (dismissed):

kberzinch (approved):

PR #2599, opened 2022-06-07
Send payment receipt exactly once

### LINKED ISSUE #2598: Payment receipt email may not have Square receipt URL in some cases
The `payment.created` event does not include the receipt URL, but that's what we're using for triggering a receipt right now, since it's exactly-once.

kberzinch:
Fix #2598

### DIFF
--- a/app/Observers/PaymentObserver.php
+++ b/app/Observers/PaymentObserver.php
@@ -40,10 +40,10 @@ class PaymentObserver
                     null !== $payment->receipt_url
                 )
             ) {
-                    $payment->receipt_sent = true;
-                    $payment->save();
+                $payment->receipt_sent = true;
+                $payment->save();
 
-                    SendPaymentReceipt::dispatch($payment);
+                SendPaymentReceipt::dispatch($payment);
             }
         });
     }
--- a/app/Observers/PaymentObserver.php
+++ b/app/Observers/PaymentObserver.php
@@ -30,21 +30,25 @@ class PaymentObserver
             SendDuesPaymentReminder::dispatch($payment->payable->user);
         }
 
-        // this is pretty cursed but i don't have a better idea on guaranteeing exactly one receipt email from
-        // ~four save events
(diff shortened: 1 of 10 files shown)


JoshOldenburg (approved):
I'm going to trust you on the `Cache::lock` logic.

kberzinch:
Idk if it's 100% correct, the docs are light on details. But i think it's close enough at least.

app/Observers/PaymentObserver.php
```diff
@@ -27,5 +29,22 @@ public function saved(Payment $payment): void
         ) {
             SendDuesPaymentReminder::dispatch($payment->payable->user);
         }
+
+        // this is pretty cursed but i don't have a better idea on guaranteeing exactly one receipt email from
+        // ~four save events
+        Cache::lock('send_payment_receipt_'.$payment->id, 5)->get(static function () use ($payment): void {
```
evan10s, line 35: The documentation about this function is lacking details and I found myself seeking out answers in the original source code.  The biggest issue is what the 5 means, I would suggest extracting it into a variable and adding a comment that it'll wait up to 5 seconds to obtain the lock ([if I understand correctly](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Cache/Lock.php#L26)).

evan10s, line 35: Do you not need to catch the `LockTimeoutException` in the event the lock can't be acquired?

kberzinch: That was also my understanding. I'll add a comment.

kberzinch: Not sure what we would do with it. Just ignore it i guess?

evan10s: Ignoring it seems fine

kberzinch: After further consideration i don't think this implementation will be a problem in practice, if it shows up in sentry I'll stick it in a try/catch

PR #2664, opened 2022-06-25
Enable DocuSign for membership agreements

### LINKED ISSUE #2516: Membership agreements on docusign

### LINKED ISSUE #2575: hasSignedLatestAgreement should be a calculated attribute instead of a method

### LINKED ISSUE #2586: Membership agreement email is sent twice

kberzinch:
Closes #2516
Closes #2575 
Closes #2586


app/Models/User.php
```diff
@@ -848,8 +849,15 @@ static function (QueryBuilder $query): void {
...
                 }
-            )
-            ->exists();
+            );
+
+        if (true === config('features.docusign-membership-agreement')) {
+            $query->whereHas('envelope', static function (Builder $query): void {
+                $query->where('complete', true);
+            });
+        }
```
JoshOldenburg, line 858: Does this handle users who signed before DocuSign? Or is it deliberate that it'll force you to re-sign?

JoshOldenburg (commented):
One quick question. Looks good overall

app/Models/User.php
kberzinch: It deliberately forces re-sign

JoshOldenburg: Sounds good.

JoshOldenburg (approved):

PR #2689, opened 2022-07-02
Add Enlightn

### LINKED ISSUE #1592: Add Enlightn to build and deploy steps
Makes Laravel-specific recommendations https://www.laravel-enlightn.com/

Best to run during CI to catch static issues and at the end of deployments to catch production configuration issues.

kberzinch:
Closes #1592


app/Http/Controllers/AttendanceReportController.php
```diff
@@ -1,36 +0,0 @@
-<?php
-
-declare(strict_types=1);
-
-namespace App\Http\Controllers;
-
-class AttendanceReportController extends Controller
```
JoshOldenburg, line 7: This was not used anywhere anymore?

config/enlightn.php
```diff
@@ -0,0 +1,208 @@
...
+    |
+    | Specify the analyzer classes that you wish to exclude from reporting. This
+    | means that if any of these analyzers fail, they will not be counted
+    | towards the exit status of the Enlightn command. This is useful
+    | if you wish to run the command in your CI/CD pipeline.
+    | Example: [\Enlightn\Enlightn\Analyzers\Security\XSSAnalyzer::class].
+    |
+    */
+    // @phan-suppress-next-line PhanPluginDuplicateConditionalUnnecessary
+    'dont_report' => true === env('SKIP_HTTP_CHECKS', false) ? [
```
JoshOldenburg, line 110: Can this just be made unconditional?

config/enlightn.php
```diff
@@ -0,0 +1,208 @@
...
+    // Format: "myorg/myrepo" like "laravel/framework".
+    'github_repo' => env('ENLIGHTN_GITHUB_REPO'),
+
+    // Set to true to restrict the max number of files displayed in the enlightn
+    // command for each check. Set to false to display all files.
+    'compact_lines' => false,
+
+    // List your commercial packages (licensed by you) below, so that they are not
+    // flagged by the License Analyzer.
+    'commercial_packages' => [
```
JoshOldenburg, line 181: Nova doesn't need to be here?

JoshOldenburg (approved):
Looks good. I have a few small questions that I assume are fine.

app/Http/Controllers/AttendanceReportController.php
kberzinch: Not that I can tell.

config/enlightn.php
kberzinch: I will be coming back to tweak this later. I need to do a full round of deployments so the nginx config updates are applied, then I can enable that check by default.

config/enlightn.php
kberzinch: Apparently not.

PR #2786, opened 2022-08-19
Remove build-time patch of subfission/cas

kberzinch:
This shouldn't be necessary since `env(...)` still returns real environment variables (just not from `.env` files), and that's how we configure the containers.


kberzinch:
Apparently the docs lied and this is still required. Once the patch is merged upstream we should be able to remove it.

PR #2952, opened 2022-10-01
Only show shirt size questions in dues flow when needed

### LINKED ISSUE #2942: Do not prompt for shirt size if not eligible for merch

JoshOldenburg:
Fixes #2942. I still need to test this.

### DIFF
--- a/resources/js/components/dues/DuesRequiredInfo.vue
+++ b/resources/js/components/dues/DuesRequiredInfo.vue
@@ -66,9 +66,9 @@
           </div>
         </div>
 
-        <h4>Information for Merchandise</h4>
+        <h4 v-if="needsSize">Information for Merchandise</h4>
 
-        <div class="form-group row">
+        <div class="form-group row" v-if="needsSize">
           <label class="col-sm-2 col-form-label">T-Shirt Size</label>
           <div class="col-sm-10 col-lg-4">
             <custom-radio-buttons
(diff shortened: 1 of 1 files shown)


resources/js/components/dues/DuesRequiredInfo.vue
```diff
@@ -274,7 +274,10 @@ export default {
       }
 
       return this.selectedPackage.restricted_to_students;
-    }
+    },
+    needsSize: function () {
+      return !this.selectedPackage || (this.merchGroupNames.length > 0) || this.$v.localUser.polo_size.$error || this.$v.localUser.shirt_size.$error;
+    },
```
evan10s, line 280: When I tested this locally, the shirt size options correctly disappeared because I'm not a student (although it's odd they appeared initially), but then reappeared when I hit next. I guess you also need to remove the validation requirement for those fields.

evan10s (changes requested):

resources/js/components/dues/DuesRequiredInfo.vue
JoshOldenburg: The visibility logic is based on the condition for the merch dropdown. I figured it's useful to have a shirt size on file for everyone. If it's not already set, I display the option. It'll disappear as soon as you select something, but I'm not super concerned about the edge case. That should only get hit for alumni who never paid dues as a student.

All that said, the implementation itself is glitchy, so I will work that out when I have time.

PR #2953, opened 2022-10-01
Allow spaces in preferred names

### LINKED ISSUE #2923: Allow spaces in preferred name field
See https://my.robojackets.org/nova/resources/users/3449 for instance

JoshOldenburg:
Fixes #2923. I still need to test this.


resources/js/components/UserEditForm.vue
```diff
@@ -245,9 +245,11 @@
 </template>
 
 <script>
-import {alpha, email, maxLength, minLength, required} from 'vuelidate/lib/validators';
+import { email, maxLength, minLength, required, helpers } from 'vuelidate/lib/validators';
 import notGTEmail from '../customValidators/notGTEmail';
 
+const alphaSpace = helpers.regex('alphaSpace', /^[a-zA-Z ]*$/);
```
evan10s, line 251: This allows there to be multiple spaces between words as well as leading and trailing spaces. I imagine we trim the input here so leading/trailing spaces aren't concerning, but can we adjust the regex to allow only one space between words?

Some testing suggests `^([a-zA-Z]+\s)*[a-zA-Z]+$` would work better (https://stackoverflow.com/a/21722204), but it's worth noting that either this or the existing proposal will block preferred names containing accents.  It's possible to write a [regex to allow accented character](https://stackoverflow.com/q/20690499)s, but not sure if we want/need to go down that rabbit hole.

Another alternative: write a looser regex to block numbers and most special characters, and replace all whitespace between words with just one space server-side

JoshOldenburg: I switched to your suggested regex. Blocking accents is deliberate. We blocked them before and I think expanding to allow them is out of scope for this PR. Additionally GT itself removes accents.

There was no server-side validation of this before. I added it.

kberzinch (changes requested):
Add some help text to indicate why the validation is failing. Right now there's just a little red icon.

image

resources/js/components/dues/DuesAdditionalInfo.vue
```diff
@@ -15,6 +15,9 @@
               id="user-preferredname"
               :class="{ 'is-invalid': $v.localUser.preferred_first_name.$error }"
               @input="$v.localUser.preferred_first_name.$touch()">
+            <div class="invalid-feedback">
+              Must be letters, with optional spaces between words only
```
kberzinch, line 19: This doesn't show up for me

kberzinch (changes requested):

JoshOldenburg:
Per Slack discussion, this doesn't 100% work but we'll merge as-is after the merge conflicts are fixed.

kberzinch (approved):

PR #3002, opened 2022-10-19
Redirect /nova/login to /login/cas

JoshOldenburg:
Nova redirects to that route intermittently, which is a 404 right now. This is probably not the _correct_ solution, but it will cover that case.


JoshOldenburg:
Per Slack discussion, I should figure out how to get the user back to where they were.

kberzinch (changes requested):
Marking to keep things tidy

kberzinch (approved):

PR #3159, opened 2022-12-15
Add new field to indicate if a merch item is distributable

### LINKED ISSUE #3073: Add field to indicate distributable Merchandise
Update the Merchandise table (and Nova) to handle the fact that not all Merchandise is actually distributable (e.g., waiving).  Probably just involves adding a column to the Merchandise table and updating Nova to use the new column instead of filtering off the word "waive."

evan10s:
This will allow us to remove some logic determining distributability based on presence of the word "waive" in a merch item's name.

Also:
* Add details about `CAS_CLIENT_SERVICE` env var to README and .env.example
* Fixed a typo (displayble -> displayable) that was present in a bunch of Nova model files
* Added a jank workaround for a Nova bug causing the `provided_by` field to not be sent to the backend by the Distribute Merchandise action

Closes #3073


database/migrations/2022_12_14_231443_add_merchandise_distributable_field.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('merchandise', static function (Blueprint $table): void {
+            $table->boolean('distributable')->default(false);
```
JoshOldenburg, line 19:
```suggestion
            $table->boolean('distributable')->default(true);
```
Per Slack disucssion

JoshOldenburg (changes requested):

kberzinch (approved):
lgtm other than note from josh

kberzinch (approved):

JoshOldenburg (approved):

PR #3345, opened 2023-02-11
Add support for international TAR forms

kberzinch:


app/Nova/Travel.php
```diff
@@ -228,6 +228,87 @@ public function fields(Request $request): array
...
+                        ->rules('required_if:is_international,1', 'nullable')
+                        ->help(
+                            'Please explain how this travel meets essential travel criteria as defined by Georgia Tech.'
+                        )
+                        ->hideFromIndex(),
+
+                    Boolean::make('Export-Controlled Technology', 'export_controlled_technology')
+                        ->required()
+                        ->help(
+                            'Do you plan to take any information or technology that is controlled?'
```
RossIV, line 251: Is the assumption that the person completing this knows whether or not something they are planning to take is subject to export controls?

app/Nova/Travel.php
```diff
@@ -228,6 +228,87 @@ public function fields(Request $request): array
...
+                        ->rules('required_if:export_controlled_technology,1', 'nullable')
+                        ->help(
+                            'If yes, please describe the information or technology.'
+                        )
+                        ->hideFromIndex(),
+
+                    Boolean::make('Embargoed Destination', 'embargoed_destination')
+                        ->required()
+                        ->help(
+                            'Do you plan to travel to an embargoed destination?'
```
RossIV, line 266: Is the assumption that the person completing this knows whether or not where they're going is an embargoed destination? Is that even allowed?

app/Nova/Travel.php
kberzinch: Short answer yes. I don't think Nova help text is the place to explain all of this.

app/Nova/Travel.php
kberzinch: Same as above. Not sure what you mean by the latter question.

RossIV: Are we/GT allowed to travel to embargoed destinations?

kberzinch: If you get a license from OFAC, sure.

RossIV (approved):

JoshOldenburg (approved):

app/Models/TravelAssignment.php
```diff
@@ -225,7 +226,11 @@ public function getTravelAuthorityRequestUrlAttribute(): string
                 'non_contract_airline' => false,
             ])['non_contract_airline']
         ) {
-            return $this->getAirfarePowerFormUrl();
+            if ($this->travel->is_international) {
+                return $this->getInternationalAirfarePowerFormUrl();
+            } else {
+                return $this->getDomesticAirfarePowerFormUrl();
+            }
```
evan10s, line 233: you could early exit here and drop the `else` braces

database/migrations/2023_02_11_172857_add_international_columns_to_travel.php
```diff
@@ -0,0 +1,50 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('travel', static function (Blueprint $table): void {
+            $table->boolean('is_international');
+            $table->boolean('export_controlled_technology');
```
evan10s, line 18: Should the booleans for things that (seem like they) are only relevant to international travel also be nullable? Seems odd to set them to a value on a domestic trip, for instance

evan10s (approved):

JoshOldenburg (approved):

PR #3407, opened 2023-02-26
Split payment and form completion emails for travel

### LINKED ISSUE #3398: Send separate emails for TAR and payment completion
There may be a significant gap between people submitting TAR forms and payments, and we should let the primary contact and treasurer know when each is completed separately.

kberzinch:
Closes #3398

View renders at https://apiary-test.robojackets.org/mailbook


database/migrations/2023_02_26_112116_split_form_and_payment_completion_emails.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('travel', static function (Blueprint $table): void {
+            $table->dropColumn('completion_email_sent');
+            $table->boolean('payment_completion_email_sent')->default(false);
+            $table->boolean('form_completion_email_sent')->default(false);
```
evan10s, line 19: This looks like it will migrate data incorrectly for existing travel assignments. I would naively suggest setting both new fields to true if `payment_completion_email_sent` is true currently

resources/views/mail/travel/allassignmentscomplete.blade.php
```diff
@@ -1,13 +1,29 @@
...
+All travelers have paid the travel fee{{ $travel->tar_required ? ' and submitted travel forms' : '' }} for {{ $travel->name }}. Contact the treasurer to book travel.
 @if($travel->tar_required)
 
-You can download all submitted forms from {{ config('app.name') }} at {{ route('nova.pages.detail', ['resource' => 'travel', 'resourceId' => $travel->id]) }}, under the action menu (three dots in top right).
+You can download all submitted forms from {{ config('app.name') }} at {{ route('nova.pages.detail', ['resource' => \App\Nova\Travel::uriKey(), 'resourceId' => $travel->id]) }}, under the action menu (three dots in top right).
 @endif
 
 Bon voyage!
+{{-- all assignments documented, not paid--}}
+@elseif($travel->assignments()->unpaid()->exists() && $travel->assignments()->needDocuSign()->doesntExist())
```
evan10s, line 12: Seems like you use these conditions quite a bit, might be worth adding a couple extra methods on the model to make the logic easier to follow

database/migrations/2023_02_26_112116_split_form_and_payment_completion_emails.php
kberzinch: I don't think it needs to be that precise. There isn't any way to trigger this email without modifying a TA, which can't happen if it's complete.

app/Http/ContentSecurityPolicy.php
```diff
@@ -56,15 +56,16 @@ public function configure(): void
             Keyword::SELF,
             Keyword::UNSAFE_INLINE,
             Keyword::REPORT_SAMPLE,
+            'https://unpkg.com',
```
JoshOldenburg, line 59: Maybe only add this when not in prod? Not sure how much it matters.

JoshOldenburg (dismissed):

app/Http/ContentSecurityPolicy.php
kberzinch: I was thinking about stripping it out at build time. Might circle back and do that later.

kberzinch: After further review, this is indeed used for some Mailbook features, though we're not using them right now. In the interest of keeping things simple I'll leave this in the CSP.

evan10s (dismissed):

PR #3478, opened 2023-03-25
Expose travel assignments for users

### LINKED ISSUE #3421: Expose travel assignments in user API

kberzinch:
Closes #3421

### DIFF
--- a/app/Http/Resources/User.php
+++ b/app/Http/Resources/User.php
@@ -11,7 +11,6 @@ use App\Http\Resources\Permission as PermissionResource;
 use App\Http\Resources\Role as RoleResource;
 use App\Http\Resources\Team as TeamResource;
 use Illuminate\Http\Request;
-use Illuminate\Http\Resources\Json\AnonymousResourceCollection;
 use Illuminate\Http\Resources\Json\JsonResource;
 use Illuminate\Support\Facades\Auth;
 
@@ -95,11 +94,7 @@ class User extends JsonResource
                     $this->resource->relationLoaded('roles') ? $this->getAllPermissions()->pluck('name') : [],
                 ]
             ),
-            'travel' => $this->when(
-                ($this->requestingSelf($request) || Auth::user()->can('manage-travel')) &&
-                $this->resource->relationLoaded('assignments'),
-                fn (): AnonymousResourceCollection => TravelAssignment::collection($this->assignments)
-            ),
+            'travel' => TravelAssignment::collection($this->whenLoaded('assignments')),
         ];
     }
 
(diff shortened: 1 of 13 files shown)


app/Models/User.php
```diff
@@ -568,6 +568,7 @@ public function getRelationshipPermissionMap(): array
             'rsvps' => 'rsvps',
             'roles' => 'roles-and-permissions',
             'permissions' => 'roles-and-permissions',
+            'assignments.travel' => 'users',
```
kberzinch, line 571: `AuthorizeInclude` assumes the permission starts with `read-`, and there's just one permission for travel called `manage-travel`. I think it's fine to expose TAs to `read-users` but I'm open to other opinions on that.

I also do want to enforce eager-loading here.

app/Http/Resources/User.php
```diff
@@ -94,6 +95,11 @@ public function toArray($request): array
                     $this->resource->relationLoaded('roles') ? $this->getAllPermissions()->pluck('name') : [],
                 ]
             ),
+            'travel' => $this->when(
+                ($this->requestingSelf($request) || Auth::user()->can('manage-travel')) &&
```
evan10s, line 99: It seems like authorize include wants `read-users` but this if statement is looking for `manage-travel`, am I missing something?

kberzinch: I guess the end result is you need both permissions. `read-users` is required to load the permission in the controller, `manage-travel` is required for it to populate in the resource here.

It's a little convoluted but I don't super want to add a `read-travel` permission just for this.

evan10s: this seems like jank that might end up confusing someone later on. I would lightly prefer (4/10) if we did something to make it more clear what's happening. AuthorizeIncludes is a bit hard to understand for newer devs or if you haven't touched the code in a while and forgot it exists

kberzinch: Fair enough. Let me play around with it a bit.

app/Traits/AuthorizeInclude.php
```diff
@@ -45,9 +45,9 @@ public function authorizeInclude(string $class, ?string $requestedInclude = null
             $permission = array_key_exists(
                 $include,
                 $relationPermMap
-            ) ? $relationPermMap[$include] : $this->camelToDashed($include);
+            ) ? $relationPermMap[$include] : 'read-'.$this->camelToDashed($include);
 
-            if (Auth::user()->cant('read-'.$permission)) {
+            if (Auth::user()->cant($permission)) {
                 Log::debug('User is missing permission: read-'.$permission);
```
evan10s, line 51: this doesn't seem to account for the new definition of `$permission` above

evan10s (approved):

app/Traits/AuthorizeInclude.php
kberzinch: Oop good catch

PR #3512, opened 2023-04-09
Replace membership agreement signing with DocuSign embedded signing

### LINKED ISSUE #2692: Clean up legacy e-sign functionality
Assuming no one yells at us about DocuSign

### LINKED ISSUE #3510: Illuminate\Database\LazyLoadingViolationException: Attempted to lazy load [user] on model [App\Models\Signature] but lazy loading is disabled.
Sentry Issue: [APIARY-BV](https://robojackets.sentry.io/issues/4067398856/?referrer=github_integration)

```
Illuminate\Database\LazyLoadingViolationException: Attempted to lazy load [user] on model [App\Models\Signature] but lazy loading is disabled.
  File "/app/Observers/MembershipAgreementTemplateObserver.php", line 27, in App\Observers\MembershipAgreementTemplateObserver::saved
    $signature->user,
  File "/app/Http/Middleware/CASAuthenticate.php", line 93, in App\Http\Middleware\CASAuthenticate::handle
    return $next($request);
  File "/app/Http/Middleware/Sentry.php", line 58, in App\Http\Middleware\Sentry::handle
    return $next($request);
  File "/public/index.php", line 52
    $request = Request::capture()
...
(83 additional frame(s) were not displayed)
```

kberzinch:
This pull request replaces both previous implementations of membership agreement signing flows with DocuSign embedded signing.

Closes #3510 
Closes #2692

Outstanding items:
- [x] Test recipient access code authentication
- [x] Switch to JWT authentication for global default sender credentials
- [x] Store completed documents within Apiary
- [x] Sender URL should have correct hostname for the envelope based on the DocuSign environment configuration
- [x] Add member name to envelope subject


app/Http/Controllers/DocuSignController.php
```diff
@@ -97,71 +142,273 @@ public function signAgreement(Request $request)
...
+                $user->docusign_refresh_token_expires_at = Carbon::now()->addDays(29);
+                $user->save();
+
+                return $this->signAgreement($request);
+        }
+    }
+
+    public function complete(Request $request)
+    {
+        if (! $request->hasValidSignatureWhileIgnoring(['event'])) {
```
kberzinch, line 383: The `event` parameter is not signed, so someone could potentially replay requests with different `event`s. I don't see a clean way to prevent that without immediately making a GET call to DocuSign, which the DocuSign docs specifically discourage.

Webhooks are too slow to rely on for interactive signing.

app/Jobs/ProcessDocuSignWebhook.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Jobs;
+
+use App\Models\DocuSignEnvelope;
+use Spatie\WebhookClient\Jobs\ProcessWebhookJob;
+
+class ProcessDocuSignWebhook extends ProcessWebhookJob
```
kberzinch, line 10: This will get more complicated over time - in particular, we're not storing the documents themselves within Apiary at this point. That's fine(ish) for membership agreements but will need to be added for travel.

app/Models/DocuSignToken.php
```diff
@@ -0,0 +1,144 @@
...
+ * @method static \Illuminate\Database\Eloquent\Builder|DocuSignToken query()
+ * @method static \Illuminate\Database\Eloquent\Builder|DocuSignToken whereCreatedAt($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|DocuSignToken whereExpiresAt($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|DocuSignToken whereToken($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|DocuSignToken whereType($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|DocuSignToken whereUpdatedAt($value)
+ *
+ * @mixin \Barryvdh\LaravelIdeHelper\Eloquent
+ */
+class DocuSignToken extends Model
```
kberzinch, line 38: This will get ripped out in favor of [JWT auth](https://developers.docusign.com/platform/auth/jwt/).

app/Jobs/ProcessDocuSignWebhook.php
kberzinch: Documents are now being received via webhook and stored on disk.

app/Http/Controllers/DocuSignController.php
```diff
@@ -97,71 +141,300 @@ public function signAgreement(Request $request)
...
     }
 
-    private static function generateAgreementPowerFormUrl(User $user): string
+    /**
+     * Handle the OAuth consent response from DocuSign.
+     *
+     * @phan-suppress PhanPluginInconsistentReturnMethod
+     * @phan-suppress PhanTypeMismatchArgumentProbablyReal
+     */
+    public function handleProviderCallback(Request $request)
```
JoshOldenburg, line 331: This really could use some docs per conversation in Slack.

JoshOldenburg (changes requested):
Documentation on the oauth flow would be very good here. It's not obvious at all from the code what's happening.

Other than that, assuming we're archiving the signed PDFs somewhere still, this looks good.

app/Mail/MembershipAgreementDocuSignEnvelopeReceived.php
```diff
@@ -30,7 +30,6 @@ public function build(): self
     {
         return $this->from('noreply@my.robojackets.org', 'RoboJackets')
             ->to($this->envelope->signedBy->gt_email, $this->envelope->signedBy->name)
-            ->cc(config('services.membership_agreement_archive_email'), 'Membership Agreement Archives')
```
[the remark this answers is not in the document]
kberzinch: This email only had the DocuSign link anyway, not the documents themselves. I'm hoping we can use a robojackets.org email address for the DocuSign service account; that will get a copy of every document sent.

JoshOldenburg (dismissed):

app/Http/Controllers/DocuSignController.php
```diff
@@ -97,71 +144,312 @@ public function signAgreement(Request $request)
...
+
+                $user->docusign_access_token = $tokens->getAccessToken();
+                $user->docusign_access_token_expires_at = Carbon::now()->addSeconds($tokens->getExpiresIn() - 60);
+                $user->docusign_refresh_token = $tokens->getRefreshToken();
+                $user->docusign_refresh_token_expires_at = Carbon::now()->addDays(29);
+                $user->save();
+
+                return $this->signAgreement($request);
+            default:
+                abort(500);
```
evan10s, line 418: Feels like we should add some kind of error/log here rather than just throwing an empty 500/

evan10s (commented):
nothing major

UX comments:
- After clicking Sign Electronically, it was unclear that I needed to click Use Company Login
- Too many exclamation points on the success popup

PR #3615, opened 2023-05-20
Add merchandise distribution endpoints

### LINKED ISSUE #3072: Add API endpoints for merchandise distribution
apiary-mobile needs some extra routes to be able to do merch distribution.

https://github.com/RoboJackets/apiary/issues/3073 should be completed before this

Necessary for merchandise->users flow:
* `GET /merchandise` - list all merch items
  * Ordered by fiscal year descending 
* `GET /merchandise/{merchandiseId}/distribute/{gtid}` - Gets distribution info of a given merchandise item for a given user and includes user shirt/polo size fields
  * Distribution info fields: Merch item metadata (id, name), distribution status (provided at, provided by)
  * Return an error if:
    * The user doesn't have a record for this merch item
    * The merchandise item is not distributable
* `POST /merchandise/{id}/distribute` - Record that a user picked up a given merch item, or error if they've already received it
  * Should accept user identification by GTID like attendance
  * Return an error if:
    * The user has already picked up this item
    * The merchandise item is not distributable

Necessary for user->merchandise flow and not immediately necessary:
*  `GET /users/{id}/merchandise` - Return all merchandise the user is or has been eligible for along with pickup info

kberzinch:
Closes #3072

### DIFF
--- a/app/Http/Controllers/MerchandiseController.php
+++ b/app/Http/Controllers/MerchandiseController.php
@@ -1,0 +1,147 @@
+<?php
+
+declare(strict_types=1);
+
+// phpcs:disable SlevomatCodingStandard.Functions.DisallowNamedArguments
+
+namespace App\Http\Controllers;
+
+use App\Http\Resources\DuesTransactionMerchandise as DuesTransactionMerchandiseResource;
+use App\Http\Resources\Merchandise as MerchandiseResource;
+use App\Http\Resources\User as UserResource;
+use App\Models\DuesTransactionMerchandise;
+use App\Models\Merchandise;
+use App\Models\User;
+use Carbon\Carbon;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Http\Request;
+use Symfony\Component\HttpFoundation\JsonResponse;
+
+class MerchandiseController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware('permission:read-merchandise|distribute-swag', ['only' => ['index']]);
+        $this->middleware('permission:distribute-swag', ['only' => ['getDistribution', 'distribute']]);
+    }
+
+    /**
+     * Display a listing of the resource.
+     */
+    public function index(): JsonResponse
+    {
+        return response()->json(
+            [
+                'status' => 'success',
+                'merchandise' => MerchandiseResource::collection(
+                    Merchandise::whereDistributable(true)
+                        ->orderByDesc('fiscal_year_id')
+                        ->get()
+                ),
+            ]
+        );
+    }
+
+    public function getDistribution(Merchandise $merchandise, User $user): JsonResponse
+    {
+        if (! $merchandise->distributable) {
+            return response()->json(
+                data: [
+                    'status' => 'error',
+                    'message' => 'item is not distributable',
+                ],
+                status: 400
+            );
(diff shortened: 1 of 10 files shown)


app/Models/User.php
```diff
@@ -585,6 +585,8 @@ public function getRelationshipPermissionMap(): array
             'roles' => 'read-roles-and-permissions',
             'permissions' => 'read-roles-and-permissions',
             'assignments.travel' => 'manage-travel',
+            'merchandise.merchandise' => 'read-merchandise',
+            'merchandise.providedBy' => 'read-users',
```
kberzinch, line 589: To eager-load all of the data, both of these keys (`merchandise.merchandise` and `merchandise.providedBy`) must be provided in the `include` parameter. It will work with just `merchandise`, but that causes N+1 queries.

I don't love this but fixing it requires a fairly substantial refactor.

JoshOldenburg (approved):

app/Http/Controllers/MerchandiseController.php
```diff
@@ -0,0 +1,147 @@
...
+        );
+    }
+
+    public function getDistribution(Merchandise $merchandise, User $user): JsonResponse
+    {
+        if (! $merchandise->distributable) {
+            return response()->json(
+                data: [
+                    'status' => 'error',
+                    'message' => 'item is not distributable',
```
evan10s, line 52: Can we make these sentence case and presentable so we can just directly show the error message in the UI?

like here `ITEM_NOT_DISTIRBUTABLE` and the app can map on its own, or the error message from the backend should be suitable to drop in the UI

app/Http/Controllers/MerchandiseController.php
```diff
@@ -0,0 +1,147 @@
...
+            );
+        }
+
+        try {
+            $dtm = self::getDuesTransactionMerchandise($merchandise, $user);
+        } catch (ModelNotFoundException) {
+            return response()->json(
+                data: [
+                    'status' => 'error',
+                    'message' => 'user does not have a paid transaction with this item',
```
evan10s, line 64: same here (maybe "No paid transaction including this item")

app/Providers/RouteServiceProvider.php
```diff
@@ -19,6 +20,9 @@ public function boot(): void
     {
         $this->configureRateLimiting();
 
+        // @phan-suppress-next-line PhanTypeMismatchReturn
+        Route::bind('gtid', static fn (string $value): User => User::whereGtid($value)->firstOrFail());
```
evan10s, line 24: what does this do?

kberzinch: This lets these two routes work with GTIDs automagically

https://github.com/RoboJackets/apiary/blob/32d503bff768f6a3fbbd770a5d5081cc985edf6f/routes/api.php#L76-L77

evan10s (approved):
I tested the API routes in test and they all seem good to me

app/Providers/RouteServiceProvider.php
evan10s: fancy

app/Models/User.php
evan10s: This causes weird server errors when these aren't specified, perhaps if you acknowledge it in a comment it'll ease someone's pain down the line

kberzinch: Idk where I would put it

PR #3682, opened 2023-06-19
Sandbox environment setup

kberzinch:
This pull request removes the previous `google-play-review` environment and replaces it with a more-general-purpose `sandbox` environment that works the same way.


evan10s (approved):
seems reasonable

Issue #3577, opened 2023-05-07
Write product and technical documentation for officers and system administrators


kberzinch:
### Scope of Work
- [x] Determine where documentation should live within GitHub
- [x] Develop a pipeline to build and deploy documentation to a user-friendly website
- [x] Write documentation for officers
    - [x] How to set up dues
    - [x] How to set up travel
    - [x] How to enter payments (including waivers)
        - [x] #2959 
    - [x] How to refund payments
    - [x] Data model
- [x] Migrate documentation that currently exists on the wiki
    - [x] [How to Accept Dues](https://wiki.robojackets.org/How_to_Accept_Dues)
    - [x] [How to Distribute Swag](https://wiki.robojackets.org/How_to_Distribute_Swag)
    - [x] [How to Record Attendance](https://wiki.robojackets.org/How_to_Record_Attendance)
    - [x] [How to Request a Self-Service Access Override](https://wiki.robojackets.org/How_to_Request_a_Self-Service_Access_Override)
    - [x] [MyRoboJackets Access Overrides](https://wiki.robojackets.org/MyRoboJackets_Access_Overrides)
    - [x] [How to fill out TAR forms](https://wiki.robojackets.org/How_to_fill_out_TAR_forms)
- [x] Write documentation for system administrators
    - [x] Deploying (a new instance of) Apiary on a new server from scratch
    - [x] Rotating compromised credentials
    - [x] Backing up an Apiary instance
    - [x] Restoring Apiary from a backup
    - [x] Managing permissions/role reference cheat sheet
    - [x] OAuth tokens
- Likely some tweaks to labels/etc to make the application more user-friendly and make the documentation more coherent

### Open Questions
- Where should documentation live within GitHub?
    - Kristaps suggests `/docs/` directory on `main` branch
- Where should documentation be deployed? 
    - Kristaps suggests similar mechanics to how we handle static assets for the application itself - could also use cloudfront/other cdn
- How can we integrate a (useful!) search box?
    - MeiliSearch? One of those frontend-only JS search engines?

PR #3777, opened 2023-07-22
Updated refund actions and documentation

kberzinch:
- Factor out Square order state check to a method on the `Payment` model and cache it for 10 seconds (with the intent of that being approximately the duration of a single request)
- Add action for refunding cash and check payments
- Refactor Square refund action
- Add documentation for refunding payments

Progress towards #3577


JoshOldenburg (changes requested):
Per Slack, add a check so users can't refund themselves. Otherwise this looks good.

JoshOldenburg (approved):

evan10s (approved):

PR #3821, opened 2023-08-06
Set up Vale

kberzinch:
Progress towards #3577


docs/officers/payments/accept.rst
```diff
@@ -1,46 +1,57 @@
...
-The following payment methods are only accepted online and incur an additional surcharge depending on the amount of the transaction.
+The following payment methods are only accepted online and incur a surcharge depending on the amount of the transaction.
 
 - Credit and debit cards
 - Google Pay
 - Apple Pay
 - Cash App
 
-We do **not** accept payments via PayPal, Venmo, or Zelle (they're more complicated and expensive to process).
+Apiary **doesn't** support payments via PayPal, Venmo, or Zelle.
```
JoshOldenburg, line 31: I prefer the old phrasing. They're not accepted because RJ doesn't want to accept them. It's not just a technical limitation. The text in the parentheses is not necessary though.

JoshOldenburg (approved):

docs/officers/payments/accept.rst
kberzinch: This was mostly an attempt to edit towards the [pronouns guidance](https://developers.google.com/style/pronouns#personal-pronouns), but I think a842b93bffba8bb0c3eb2560408329cc5d5ef73f is the clearest option.

docs/admins/access-overrides.rst
```diff
@@ -1,51 +1,61 @@
...
-Required Tasks
+Required tasks
 ~~~~~~~~~~~~~~
 
 - Sign the membership agreement
 - Attend a team meeting
-   - Attendance must be recorded in MyRoboJackets for it to count.
-   - Only teams in MyRoboJackets that have the :guilabel:`Self-Service Override Eligible` property set to true in Nova will count towards this requirement.
-     This means that event attendance (e.g., General Interest meetings), training attendance, and Core or discipline Core attendance will not satisfy this requirement.
+   - An officer must record attendance in Apiary for it to count.
```
evan10s, line 47: I don't like the use of "officer" here, especially for admin-facing docs, we should be more specific about the exact requirements. something like "either directly or via a remote attendance link"

evan10s (dismissed):

PR #4201, opened 2023-12-24
Swap travel PowerForms for DocuSign API integration

### LINKED ISSUE #3578: Update travel forms DocuSign workflow to a direct API integration
### Scope of Work
- Replace the existing PowerForms workflow with a direct DocuSign API integration
- Refactor templates to individual documents that can be composed at the time of sending (current implementation is ~4 different templates with different combinations of forms)
- #2674 

### Open Questions
- The forms we collect are owned by Procurement & Business Services, so we cannot own templates for those forms. Do we need to convince PBS to build templates?
    - Alternate option: don't use templates, build the entire envelope definition from scratch (including PDF upload, form field setup, etc)
- Should documents be sent by the service account or the primary contact for the trip?
- Are we even supposed to still be collecting Travel Authority Request forms? They're not listed on PBS website anymore

kberzinch:
This pull request closes #3578.

This is not the same flow as #3512 for various reasons - there are no travel templates in DocuSign, Apiary will upload PDFs and configure all form fields on a per-envelope basis, and we are using the trip's primary contact as the sender instead of the service account.

- All legacy PowerForm code is removed
- Envelopes are triggered automatically when a travel assignment is created
    - Traveler must have a DocuSign account, a phone number, and emergency contact information
    - Primary contact must have valid DocuSign credentials
- `/sign/travel` now goes through DocuSign embedded signing similar to membership agreements
- `/sign/auth/user` now allows refreshing DocuSign credentials on-demand (needed for trip primary contacts)

Deliberately left out of scope:
- Airfare-specific fields will be populated separately after #4209 is merged
- Admin-facing changes will be covered as part of other Nova work for travel


app/Observers/TravelAssignmentObserver.php
```diff
@@ -6,15 +6,22 @@
...
 use App\Jobs\SendTravelAssignmentReminder;
 use App\Models\TravelAssignment;
-use App\Notifications\Travel\TravelAssignmentCreated;
 
 class TravelAssignmentObserver
 {
     public function created(TravelAssignment $assignment): void
     {
-        $assignment->user->notify(new TravelAssignmentCreated($assignment));
+        if ($assignment->travel->tar_required === true || $assignment->travel->needs_airfare_form) {
```
kberzinch, line 18: Add else case if no DocuSign forms are required

app/Http/Controllers/DocuSignController.php
```diff
@@ -254,6 +304,28 @@ public function handleProviderCallback(Request $request)
...
+
+                /** @var \App\Models\User $user */
+                $user = $request->user();
+
+                if ($userinfo->getEmail() !== $user->uid.'@gatech.edu') {
+                    $userinfo_serialized = DocuSign::serializeUserInfo($userinfo);
+
+                    Log::info($userinfo_serialized);
+
+                    abort(401);
```
evan10s, line 319: aborting with a plain 401 should probably at least log a message saying what happened, rather than the user info with no context

app/Http/Controllers/DocuSignController.php
```diff
@@ -254,6 +304,28 @@ public function handleProviderCallback(Request $request)
...
+                }
+
+                $user->docusign_access_token = $tokens->getAccessToken();
+                $user->docusign_access_token_expires_at = Carbon::now()->addSeconds($tokens->getExpiresIn() - 60);
```
evan10s, line 323: nit: I prefer to make immutable dates using `CarbonImmutable`

app/Http/Controllers/DocuSignController.php
```diff
@@ -288,6 +360,28 @@ public function handleProviderCallback(Request $request)
...
+                // In this case, a user has just authenticated with DocuSign after starting the signing flow.
+                // We need to store the credentials in their user model, then continue with signing.
+
+                /** @var \App\Models\User $user */
+                $user = $request->user();
+
+                if ($userinfo->getEmail() !== $user->uid.'@gatech.edu') {
+                    $userinfo_serialized = DocuSign::serializeUserInfo($userinfo);
+
+                    Log::info($userinfo_serialized);
```
evan10s, line 373: same here re logging

app/Jobs/ProcessDocuSignWebhook.php
```diff
@@ -112,6 +112,14 @@ public function handle(): void
...
+                        $envelope->travel_authority_filename = $disk_path;
+                    }
+
+                    if (str_contains($document['name'], 'Airfare')) {
+                        $envelope->direct_bill_airfare_filename = $disk_path;
+                    }
+
                     if ($document['type'] === 'summary') {
                         $envelope->summary_filename = $disk_path;
                     }
```
evan10s, line 125: should these be `elseif`s?

app/Util/DocuSign.php
```diff
@@ -366,4 +381,421 @@ public static function membershipAgreementEnvelopeDefinition(DocuSignEnvelope $e
...
+
+    private static function emailSubject(Travel $travel): string
+    {
+        if ($travel->tar_required && $travel->needs_airfare_form) {
+            return $travel->name.' Travel Forms';
+        } elseif ($travel->tar_required && ! $travel->needs_airfare_form) {
+            return $travel->name.' Travel Information Form';
+        } elseif (! $travel->tar_required && $travel->needs_airfare_form) {
+            return $travel->name.' Airfare Request Form';
+        } else {
```
evan10s, line 393: nit: `else` block is unnecessary here

app/Util/DocuSign.php
```diff
@@ -366,4 +381,421 @@ public static function membershipAgreementEnvelopeDefinition(DocuSignEnvelope $e
...
+    /**
+     * Builds the array of documents for this travel assignment.
+     *
+     * @return array<\DocuSign\eSign\Model\Document>
+     */
+    private static function travelAssignmentDocuments(TravelAssignment $assignment): array
+    {
+        $documents = [];
+
+        if ($assignment->travel->tar_required === true) {
```
evan10s, line 736: is there a reason to leave `=== true` here?

```suggestion
        if ($assignment->travel->tar_required) {
```

app/Util/DocuSign.php
```diff
@@ -366,4 +381,421 @@ public static function membershipAgreementEnvelopeDefinition(DocuSignEnvelope $e
...
+            ->setMessageLock(true)
+            ->setEmailSettings(self::travelAssignmentEmailSettings($envelope->signable))
+            ->setNotification(
+                (new Notification())
+                    ->setUseAccountDefaults(false)
+                    ->setReminders(
+                        (new Reminders())
+                            ->setReminderEnabled(true)
+                            ->setReminderDelay(1)
+                            ->setReminderFrequency(1)
```
evan10s, line 779: what is the unit here?

app/Util/DocuSign.php
kberzinch: `tar_required` is technically nullable

app/Util/DocuSign.php
kberzinch: Days

evan10s:
Went through _a_ travel assignment flow with DocuSign and didn't notice anything glaring

evan10s (approved):

JoshOldenburg (approved):

PR #4209, opened 2023-12-27
Add support for Matrix itineraries for airfare requests

### LINKED ISSUE #4200: Require Matrix itinerary for flights
Validate carrier

Validate fare is round trip or one way originating from ATL

Validate price is within allowed range for trip fee

Validate legs are nonstop

Some kind of override / debug mode

kberzinch:
This pull request closes #4200.

Remaining tasks:
- [x] Allow one-way flights by default


app/Nova/Travel.php
```diff
@@ -370,6 +402,31 @@ public function cards(Request $request): array
...
     }
 
+    /**
+     * Handle any post-validation processing.
+     *
+     * @param  \Illuminate\Validation\Validator  $validator
+     */
+    protected static function afterValidation(NovaRequest $request, $validator): void
+    {
+        $totalCost = $request->tar_lodging + $request->tar_registration;
```
kberzinch, line 412: Check airfare price here

app/Models/Airport.php
```diff
@@ -0,0 +1,72 @@
...
+        foreach ($raw as $details) {
+            if ($details['iata'] === '' || $details['iata'] === null) {
+                continue;
+            }
+
+            $rows[] = [
+                'iata' => strtoupper($details['iata']),
+                'name' => $details['name'],
+                'city' => $details['city'],
+            ];
```
kberzinch, line 57: Need to add country code here for detecting international destinations

app/Util/BusinessTravelPolicy.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Util;
+
+class BusinessTravelPolicy
+{
+    public const TRIP_FEE_COST_RATIO = 0.2;
```
JoshOldenburg, line 9: Shouldn't this be in config? Also, why 20%?

app/Nova/Actions/MatrixAirfareSearch.php
```diff
@@ -0,0 +1,212 @@
...
+                ->required(),
+
+            BooleanGroup::make('Policy Filters')
+                ->options(MatrixItineraryBusinessPolicy::POLICY_LABELS)
+                ->default(static fn (): array => $trip->airfare_policy ?? [])
+                ->rules('required', new FareClassPolicyRequiresMarketingCarrierPolicy())
+                ->required()
+                ->help(
+                    'Select policies to apply as search filters. <strong>Note that selected itineraries must still '.
+                    'meet the policy configured for the trip.</strong>'
```
JoshOldenburg, line 200: The way this doesn't override the `Travel`'s policy isn't 100% clear but I don't have a suggestion.

app/Nova/TravelAssignment.php
```diff
@@ -93,6 +101,36 @@ public function fields(NovaRequest $request): array
                             ?->id
                 ),
 
+            Code::make('Matrix Itinerary')
```
JoshOldenburg, line 104: It's a little awkward how the field defaults to `null` on the TA creation page. Not a deal breaker.

app/Nova/TravelAssignment.php
```diff
@@ -93,6 +101,36 @@ public function fields(NovaRequest $request): array
...
+                ->json()
+                ->rules('nullable', 'json', new MatrixItineraryDataStructure())
+                ->required()
+                ->help(
+                    'If this trip includes airfare, you must provide an itinerary in Matrix JSON format. Search '.
```
JoshOldenburg, line 109: It would be slick to add the `required` rule dynamically if the `Travel` has airfare enabled. Not a deal breaker.

JoshOldenburg (commented):
As discussed in Slack, when users without `update-airfare-policy` (or so we're guessing) create a `Travel`, all of the policies are unchecked. This is being deliberately ignored for now.

Couple questions before approving it.

Overall I'm a big fan of this. The UX/flow is awkward in some places but with no obvious way to improve it other than good docs. I worry the action will get forgotten.

app/Nova/Actions/MatrixAirfareSearch.php
```diff
@@ -0,0 +1,212 @@
...
+                ->rules('required')
+                ->required(),
+
+            Tag::make('Destination Airports', 'destination_airports', Airport::class)
+                ->rules('required')
+                ->required(),
+
+            Number::make('Passenger Count')
+                ->min(1)
+                ->max(9)
```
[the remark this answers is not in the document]
kberzinch: That's the most Matrix supports.

app/Nova/Travel.php
```diff
@@ -104,8 +111,8 @@ public function fields(Request $request): array
             Currency::make('Fee', 'fee_amount')
                 ->sortable()
                 ->required()
-                ->rules('required', 'integer')
-                ->min(10)
+                ->rules('required', 'integer', 'min:20', 'max:1000')
+                ->min(20)
```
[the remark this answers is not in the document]
kberzinch: No, it's arbitrary. The lowest fee that's been collected so far is $17.

app/Util/BusinessTravelPolicy.php
kberzinch: Also arbitrary

app/Nova/TravelAssignment.php
kberzinch: Yeah I fiddled around with that but I couldn't get it to be blank.

app/Nova/TravelAssignment.php
kberzinch: Yep, I'll handle that when I redo TA creation.

app/Util/BusinessTravelPolicy.php
JoshOldenburg: The travel policy wiki page suggests 10% so PMs might be confused. imo leave it lower and let that be a management decision instead of a software-enforced decision.

kberzinch: Yeah, I know. I've had conversations with Rahul and I think the policy this year will be higher than 10%, but I'll leave it configurable.

JoshOldenburg (approved):

app/Nova/TravelAssignment.php
```diff
@@ -93,6 +100,36 @@ public function fields(NovaRequest $request): array
...
+                    'for flights at <a href="https://matrix.itasoftware.com">Matrix</a>, click <strong>Copy itineray '.
```
evan10s, line 109:
```suggestion
                    'for flights at <a href="https://matrix.itasoftware.com">Matrix</a>, click <strong>Copy itinerary '.
```

database/migrations/2023_12_26_174218_add_update_travel_policy_permission.php
```diff
@@ -0,0 +1,28 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        app()['cache']->forget('spatie.permission.cache');
+
```
evan10s, line 16: this line feels unnecessary

evan10s:
I put a trip start date in 2025 and managed to get an error from Matrix about it:
image

database/migrations/2023_12_26_174218_add_update_travel_policy_permission.php
evan10s: sorry 😬 I only meant the extra line break, I believe the cache clearing is important here?

evan10s:
Matrix itinerary field defaults to `null` when adding the first travel assignment for a trip:
image

database/migrations/2023_12_26_174218_add_update_travel_policy_permission.php
kberzinch: I'll revert the whole commit. I prefer the whitespace and the style checkers don't have a problem with it :wink:

The cache clear _probably_ is not necessary since it's already cleared before migrations run, but I've been adding it to every permissions migration

kberzinch:
> Matrix itinerary field defaults to `null` when adding the first travel assignment for a trip

This is a Nova limitation, I can't do anything about it. (It also applies to every assignment, not just the first one)

kberzinch:
> I put a trip start date in 2025 and managed to get an error from Matrix about it
Will add validation on trip date fields and action date fields

evan10s (dismissed):

resources/views/travel/matrixitinerarypreview.blade.php
```diff
@@ -0,0 +1,18 @@
...
+@foreach($slice['segments'] as $segment)
+<div style="display: grid; grid-template-columns: 5% 95%">
+    <img alt="{{ $segment['carrier']['shortName'] }}" src="https://www.gstatic.com/flights/airline_logos/35px/{{ $segment['carrier']['code'] }}.png" style="align-self: center;">
+    <div style="display: grid; grid-template-columns: 50% 25% 25%;">
+        <div style="display: block;"><strong>{{ $segment['origin']['city']['name'] }} ({{$segment['origin']['code']}}) to {{ $segment['destination']['city']['name'] }} ({{ $segment['destination']['code'] }})</strong> on {{ \Carbon\Carbon::parse($segment['departure'])->format('D, M j') }}</div>
+        <div style="display: block; grid-column: 2/span 2;"><strong>{{ \Carbon\Carbon::parse($segment['departure'])->format('g:i A') }}</strong> to <strong>{{ \Carbon\Carbon::parse($segment['arrival'])->format('g:i A') }}</strong> ({{ \Carbon\Carbon::parse($segment['arrival'])->diffForHumans(\Carbon\Carbon::parse($segment['departure']), \Carbon\CarbonInterface::DIFF_ABSOLUTE, true, 2) }})</div>
```
evan10s, line 8: can we render the year here too (also for arrival date)?

database/migrations/2023_12_26_174218_add_update_travel_policy_permission.php
evan10s: fair

resources/views/travel/matrixitinerarypreview.blade.php
kberzinch: Not sure what you meant by arrival date but I added the year to the two places we render a date here

app/Nova/TravelAssignment.php
```diff
@@ -199,9 +200,8 @@ static function (string $message) use ($validator, &$businessPolicyPassed): void
             if ($trip->fee_amount / $total_cost < config('travelpolicy.minimum_trip_fee_cost_ratio')) {
                 $validator->errors()->add(
                     'matrix_itinerary',
-                    'Trip fee must be at least '.
-                    (config('travelpolicy.minimum_trip_fee_cost_ratio') * 100).
-                    '% of the per-person cost for this trip.'
+                    'The airfare cost exceeds the amount allowed for this trip. Increase the trip fee to at '.
+                    'least 20% of the per-person total cost for this trip.'
```
evan10s, line 204: is this intentionally hard-coded as `20%`?

kberzinch: Good catch

evan10s:
image

I believe I have admin on apiary-test, but this field is still read-only for me. The help text makes it sound like my access would allow me to change it

resources/views/travel/matrixitinerarypreview.blade.php
evan10s: oops, yes, the return segment departure date. looks correct to me in Nova

PR #4250, opened 2024-01-07
Adding a human readable graduation semester for user profiles in Nova

jvogt23:
Adds a field into Nova under the Users category for graduation semesters. This field necessitates a valid semester value in the MySQL database the information is pulled from, so graduation semester entries are now validated server-side using a regular expression.


app/Nova/User.php
```diff
@@ -809,4 +815,31 @@ private static function adminCanRun(NovaRequest $request): bool
     {
         return $request->user()->hasRole('admin');
     }
+
+    //Helper method: Graduation semester is a 6-digit code by default.
+    //Prepares semester data for display
```
kberzinch, line 820: Use docblock format

app/Nova/User.php
```diff
@@ -809,4 +815,31 @@ private static function adminCanRun(NovaRequest $request): bool
...
+    private static function parseGradSemester(?string $sem): string
+    {
+        if ($sem === null) {
+            return '';
+        }
+        if (preg_match('/^[0-9]{4}0[258]$/', $sem) === false) {
```
kberzinch, line 826: `preg_match` only returns `false` if there is an _error_. If the regex is valid but does not match the input string, it returns `0`. [[docs](https://www.php.net/manual/en/function.preg-match.php)]

app/Nova/User.php
```diff
@@ -809,4 +815,31 @@ private static function adminCanRun(NovaRequest $request): bool
...
+            throw new InvalidArgumentException('Invalid date code for field \'graduation_semester\'.');
```
kberzinch, line 827: Return `null` or empty string here. This field is not critical; if it's not valid it's better to just leave it blank rather than throwing an error.

app/Nova/User.php
```diff
@@ -809,4 +815,31 @@ private static function adminCanRun(NovaRequest $request): bool
...
+        }
+        $semcode = substr($sem, 4);
+        $output = '';
+        switch ($semcode) {
```
kberzinch, line 831: Add default case to return `null` or empty string.

kberzinch (changes requested):
Just a few small notes. Looks like it works OK in sandbox.

jvogt23:
After updating the branch, there are now two "action required" checkstyle checks. Before the update, my changes passed all checks.

kberzinch (approved):

evan10s:
I can do a more thorough review tonight, but can you change the PR title from the default to something more descriptive?

JoshOldenburg (approved):
Looks good to me

app/Nova/User.php
```diff
@@ -809,4 +813,36 @@ private static function adminCanRun(NovaRequest $request): bool
     {
         return $request->user()->hasRole('admin');
     }
+
+    /**
+     * Helper method: Graduation semester is a 6-digit code by default.
+     * Prepares semester data for display, returning empty if no string/invalid
+     * string is entered.
+     */
+    private static function parseGradSemester(?string $sem): string
```
evan10s, line 822: This code might be better placed as an attribute on the user model? as in, add a method called `getHumanReadableGraduationSemester` (unsure what a good name is, but the attribute value would be `Fall 2023`, etc. like you do here)

here's an example: https://github.com/RoboJackets/apiary/blob/b6b74b15383f419e9fed6de876bb6da4a357ca7d/app/Models/User.php#L397-L402

evan10s, line 822: I would gently just not abbreviating semester `$sem` here and naming the variable `$semester` so it's easier for other people to follow later

app/Nova/User.php
```diff
@@ -809,4 +813,36 @@ private static function adminCanRun(NovaRequest $request): bool
...
+    {
+        if ($sem === null) {
+            return '';
+        }
+        if (preg_match('/^[0-9]{4}0[258]$/', $sem) === 0 || preg_match('/^[0-9]{4}0[258]$/', $sem) === false) {
+            return '';
+        }
```
evan10s, line 829: - PHP shouldn't evaluate later parts of an OR conditional if the first is true, so you can merge this `if` statement with the previous null check
- I would pull out the `preg_match` here since it's called identically twice; it'll make the condition more compact

app/Nova/User.php
```diff
@@ -809,4 +813,36 @@ private static function adminCanRun(NovaRequest $request): bool
...
+                $output .= 'Spring ';
+                break;
+            case '05':
+                $output .= 'Summer ';
+                break;
+            default:
+                return '';
+        }
+
+        return $output.substr($sem, 0, 4);
```
evan10s, line 846: Won't this cut out the season that was just concatenated onto output? it may be easier to read this if we ended by concatenating the `$year` and `$season`.

app/Nova/User.php
kberzinch: I think that example is considered legacy now; [the Laravel docs have a different syntax](https://laravel.com/docs/10.x/eloquent-mutators#accessors-and-mutators). Same idea though.

app/Models/User.php
```diff
@@ -570,6 +570,38 @@ public function getRememberTokenName(): string
...
+     * Graduation Semester is a 6-digit code by default.
+     * To retrieve a more readable form, this method obtains the
+     * code for this User and translates it to the format "Fall xxxx.
+     */
+    public function getHumanReadableSemester(): string
+    {
+        $semester = $this->graduation_semester;
+        if ($semester === null || preg_match('/^[0-9]{4}0[258]$/', $semester) === 0) {
+            return '';
+        } //Testing
```
evan10s, line 583: comment says `Testing`, did you mean to leave this in?

app/Models/User.php
```diff
@@ -570,6 +570,38 @@ public function getRememberTokenName(): string
         throw new BadMethodCallException('Not implemented');
     }
 
+    /**
+     * Graduation Semester is a 6-digit code by default.
+     * To retrieve a more readable form, this method obtains the
+     * code for this User and translates it to the format "Fall xxxx.
+     */
+    public function getHumanReadableSemester(): string
```
evan10s, line 578: if you append `Attribute` to this function name, laravel will magically turn this into a User model attribute called `human_readable_semester`

app/Models/User.php
```diff
@@ -570,6 +570,38 @@ public function getRememberTokenName(): string
         throw new BadMethodCallException('Not implemented');
     }
 
+    /**
+     * Graduation Semester is a 6-digit code by default.
+     * To retrieve a more readable form, this method obtains the
+     * code for this User and translates it to the format "Fall xxxx.
```
evan10s, line 576:
```suggestion
     * code for this User and translates it to the format "[season] yyyy"
```

evan10s (commented):
a few more small things, otherwise lgtm

evan10s (approved):

app/Models/User.php
```diff
@@ -570,6 +570,38 @@ public function getRememberTokenName(): string
         throw new BadMethodCallException('Not implemented');
     }
 
+    /**
+     * Graduation Semester is a 6-digit code by default.
+     * To retrieve a more readable form, this method obtains the
+     * code for this User and translates it to the format "[season] yyyy".
+     */
+    public function getHumanReadableSemesterAttribute(): string
```
kberzinch, line 578:
```suggestion
    public function getHumanReadableGraduationSemesterAttribute(): string
```

app/Nova/User.php
```diff
@@ -171,6 +171,13 @@ public function fields(NovaRequest $request): array
                 ->onlyOnDetail()
                 ->hideFromDetail(static fn (NovaRequest $r, AppModelsUser $u): bool => $u->is_service_account),
 
+            Text::make(
+                'Graduation Semester',
+                static fn (AppModelsUser $u): string => $u->getHumanReadableSemesterAttribute()
+            )
```
kberzinch, line 177: Because this is now a computed attribute on the model, it does not need to be a computed field on the Nova side.

```suggestion
            Text::make('Graduation Semester', 'human_readable_graduation_semester')
```

kberzinch (changes requested):

kberzinch (approved):
Please update the PR title

PR #4278, opened 2024-01-13
Finish DocuSign integration for travel forms

kberzinch:
- DocuSign PDFs updated with cleaned documents
- Travel packet includes a detailed itinerary if airfare is requested
- DocuSign travel emails are included in Mailbook
- New fields required for trip forms added to the `Travel` model and Nova


app/Models/Airport.php
```diff
@@ -54,7 +74,7 @@ public function getRows(): array
                 'iata' => strtoupper($details['iata']),
                 'name' => $details['name'],
                 'city' => $details['city'],
-                'state' => $details['state'],
+                'state' => str_replace('-', ' ', $details['state']),
```
kberzinch, line 77: For some reason, this dataset has dashes where there should be spaces. This fixes that.

routes/mailbook.php
```diff
@@ -2036,3 +2037,172 @@ static function () use ($otherUser, $travel): TravelAssignment {
...
+                'travel_id' => $travel->id,
+            ]);
+            $assignment->save();
+
+            return $assignment;
+        });
+
+        return new TravelMail($assignment);
+    })
+    ->variant('Only Airfare Request Form', static function (): TravelMail {
```
kberzinch, line 2097: This is not possible to trigger with the current Nova setup, but it is supported on the backend, and it will be an option after the next PR.

JoshOldenburg (approved):

evan10s (approved):

PR #4403, opened 2024-02-17
Prune access from access-inactive users

### LINKED ISSUE #3695: Revoke elevated access from access-inactive users
Daily job to remove any attached roles and permissions from users that are not access active

Non-member role is already reattached automatically if needed 

Also remove PM relationship from teams

Sync access to Jedi when done

kberzinch:
Closes #3695

### DIFF
--- a/app/Jobs/PruneAccessFromAccessInactiveUsers.php
+++ b/app/Jobs/PruneAccessFromAccessInactiveUsers.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Jobs;
+
+use App\Models\User;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+
+class PruneAccessFromAccessInactiveUsers implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    /**
+     * Create a new job instance.
+     */
+    public function __construct()
+    {
+        $this->queue = 'jedi';
+    }
+
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+        User::accessInactive()
+            ->where('is_service_account', '=', false)
+            ->where(static function (Builder $query): void {
+                $query->whereHas('roles')
+                    ->orWhereHas('permissions');
+            })
+            ->get()
+            ->each(static function (User $user): void {
+                $user->roles()->detach();
+                $user->permissions()->detach();
+
+                $user->manages()
+                    ->update(['project_manager_id' => null]);
+
+                $user->novaNotifications()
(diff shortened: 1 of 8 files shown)


app/Jobs/PruneAccessFromAccessInactiveUsers.php
```diff
@@ -0,0 +1,57 @@
...
+                $user->roles()->detach();
+                $user->permissions()->detach();
+
+                $user->manages()
+                    ->update(['project_manager_id' => null]);
+
+                $user->novaNotifications()
+                    ->delete();
+
+                $user->searchable();
```
JoshOldenburg, line 52: What does this do?

kberzinch: Reindexes the user in Meilisearch (since some places use role filters in Meilisearch)

evan10s (commented):
Given that if access ends unintentionally early, it would create a decent amount of human effort to restore all PMs/trainer/team/admin access in apiary, we should probably add a grace period in this job.

JEDI doesn't wait, but its restoration process is automated. I think we should add at least a 24 hour grace period before removing anything from Apiary.

Also, can we disable this behavior on a config flag for non-prod environments? It would be annoying to have to keep up access active statuses or manually reset access in tinker for that.

kberzinch:
I'll add a feature flag. Are you OK with merging this in combination with #4408?

evan10s:
> I'll add a feature flag. Are you OK with merging this in combination with #4408?

Even with the notification beforehand, I think the grace period is still prudent

evan10s (approved):

JoshOldenburg (commented):
I don't see where you're actually using the new timestamp logic. All I see is the `! $user->is_access_active` in the job. Am I missing something?

app/Jobs/PruneAccessFromAccessInactiveUsers.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $this->queue = 'jedi';
+    }
+
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+        User::accessInactive(CarbonImmutable::now()->subDay())
```
kberzinch, line 36: The timestamp passed in here is 1 day ago, so the DB results will contain users that were inactive as of 1 day ago

app/Jobs/PruneAccessFromAccessInactiveUsers.php
```diff
@@ -0,0 +1,59 @@
...
+            ->where('is_service_account', '=', false)
+            ->where(static function (Builder $query): void {
+                $query->whereHas('roles')
+                    ->orWhereHas('permissions');
+            })
+            ->get()
+            ->filter(static fn (User $user): bool => ! $user->is_access_active)
```
kberzinch, line 43: The check down here is to make sure the users are _still_ access-inactive. I can't easily chain the same scope on itself so this seemed like the least bad option.

kberzinch (commented):
See below inline comments

app/Jobs/PruneAccessFromAccessInactiveUsers.php
JoshOldenburg: This is the line I was missing. Thanks.

JoshOldenburg (approved):

PR #4519, opened 2024-04-01
Added RSVPs and Attendance cards to main dashboards

jvogt23:
Title.

###Expected Behavior
This change allows Nova users to see the RSVPs and attendance for each upcoming event. Events will not have information shown on the main dashboard if nobody has RSVPed or attended the event yet, and only upcoming events are shown (Any event whose end date is after Carbon::now() will not be shown).

###Files changed/Files to check
App\Nova\Dashboards\Main
App\Nova\Metrics\ActiveAttendanceBreakdown
App\Nova\Metrics\RsvpSourceBreakdown


kberzinch (changes requested):
Rebase from master

app/Nova/Metrics/ActiveAttendanceBreakdown.php
```diff
@@ -81,6 +113,7 @@ public function calculate(Request $request): PartitionResult
      */
     public function uriKey(): string
     {
-        return 'active-attendance-breakdown';
+        return $this->resourceId === -1 ? 'active-attendance-breakdown' :
+            'active-attendance-breakdown-'.$this->attendableType.'-'.$this->resourceId;
```
kberzinch, line 117: Playing around with this locally and I think this is a little bit cleaner
```suggestion
        return $this->resourceId === -1 ? 'active-attendance-breakdown' :
            '../'.$this->attendableType.'/'.$this->resourceId.'/metrics/active-attendance-breakdown';
```

That way you don't need the card-specific routing logic on the dashboard level, because the URI actually points to the event-specific metric.

app/Nova/Metrics/ActiveAttendanceBreakdown.php
```diff
@@ -27,20 +33,42 @@ public function name(): string
...
+    /**
+     * If displaying on the main dashboard, this indicates the event to get attendance from.
+     */
+    private int $resourceId = -1;
+
+    /**
+     * If displaying on a page different from the type of attendable you are showing.
+     * This indicates the morphClass of the attendable (for example, an event).
+     */
+    private string $attendableType;
```
kberzinch, line 45: Make both of these nullable and default to null, so that you can use null coalescing further down - https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op

app/Nova/Metrics/ActiveAttendanceBreakdown.php
```diff
@@ -27,20 +33,42 @@ public function name(): string
...
     {
+        $resourceId = -1;
+        $attach_to_resource_id = false;
+        if ($this->resourceId !== -1) {
+            $resourceId = $this->resourceId;
+            $attach_to_resource_id = true;
+        } elseif (isset($request->resourceId)) {
+            $resourceId = $request->resourceId;
+            $attach_to_resource_id = true;
+        }
```
kberzinch, line 71:
```suggestion
        $resourceId = $request->resourceId ?? $this->resourceId;
```

app/Nova/Metrics/ActiveAttendanceBreakdown.php
```diff
@@ -50,10 +78,14 @@ public function calculate(Request $request): PartitionResult
                 'active'
             );
 
-        if (isset($request->resourceId)) {
+        if ($attach_to_resource_id) {
```
kberzinch, line 81:
```suggestion
        if ($resourceId !== null) {
```

app/Nova/Metrics/ActiveAttendanceBreakdown.php
```diff
@@ -50,10 +78,14 @@ public function calculate(Request $request): PartitionResult
...
             $query = $query
-                ->where('attendable_id', $request->resourceId)
-                ->where('attendable_type', $request->model()->getMorphClass());
+                ->where('attendable_id', $resourceId)
+                ->where(
+                    'attendable_type',
+                    $this->attendableType !== '' ? $this->attendableType :
+                    $request->model()->getMorphClass()
+                );
```
kberzinch, line 88:
```suggestion
                ->where('attendable_id', $resourceId)
                ->where(
                    'attendable_type',
                    $request?->model()?->getMorphClass() ?? $this->attendableType
                );
```

app/Nova/Metrics/RsvpSourceBreakdown.php
```diff
@@ -4,28 +4,47 @@
...
+        )->sole()->name;
+    }
+
+    /**
+     * The resource ID attached to this metric.
+     * Used when the resource ID cannot be inferred from the Nova page.
+     */
+    protected int $resourceId;
+
+    public function __construct(int $resourceId = -1)
```
kberzinch, line 33: Same comments as for `ActiveAttendanceBreakdown`

app/Nova/User.php
```diff
@@ -112,6 +113,10 @@ class User extends Resource
     public function fields(NovaRequest $request): array
     {
         return [
+            Hidden::make('Has Ever Logged In')
+                ->showOnCreating()
+                ->default(static fn (Request $r) => 0),
```
kberzinch, line 118: Nitpick - this should be a boolean attribute
```suggestion
                ->default(static fn (Request $r): false => false),
```

kberzinch (changes requested):

app/Nova/Metrics/ActiveAttendanceBreakdown.php
jvogt23: The issue with this suggestion is that the ActiveAttendanceBreakdown doesn't always attach to a resource ID. I attempted to do the resource ID things in the way you suggested, and it broke the card for me.

app/Nova/Event.php
```diff
@@ -81,6 +81,8 @@ class Event extends Resource
     public function fields(Request $request): array
     {
         return [
+            Text::make('Event ID', 'id')
+                ->onlyOnDetail(),
```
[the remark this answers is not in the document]
jvogt23: Ah, didn't catch that. That is a debug thing.

app/Nova/Metrics/ActiveAttendanceBreakdown.php
jvogt23: Unable to make this work - Will try after a different card, and if does not work, will notify

app/Nova/Metrics/ActiveAttendanceBreakdown.php
jvogt23: Not working for now - Will revisit after trying a few other cards

app/Nova/Metrics/ActiveAttendanceBreakdown.php
kberzinch: They generally would have an 's' appended though I think there's a more robust way to do it. I would just append it for now.

There are only two types of attendables, events and teams, so that should be fine.

jvogt23: confirmed, this works - also the uri actually starts with nova-api, not nova/dashboards, for the record

kberzinch (approved):

Issue #2749, opened 2022-08-01
Add support for Mobile Credential for kiosk


kberzinch:
Mobile app support is out of scope but should be straightforward once we have it working on the kiosk.

PR #4594, opened 2024-05-04
Support access card numbers for attendance

kberzinch:
Backend changes to eventually support #2749

- Adds `AccessCard` model to track `gtAccessCardNumber`s associated with `User`s
- Pulls `gtAccessCardNumber` from CAS and BuzzAPI if available, and associates the number with the local `User`
    - Existing attendance records with this access card number will be backfilled with the matching GTID
- Supports receiving `access_card_number` in addition to or instead of `gtid` when creating attendance records, and will look up the GTID to add to the attendance record if available

Note that BuzzAPI does not currently support looking up accounts by `gtAccessCardNumber`s, so we will depend on other events to trigger the lookup.


config/cas.php
```diff
@@ -176,15 +176,16 @@
...
+    'cas_masquerade' => env('CAS_MASQUERADE'),
+    'cas_masquerade_gtGTID' => env('CAS_MASQUERADE_gtGTID'),
+    'cas_masquerade_email_primary' => env('CAS_MASQUERADE_email_primary'),
+    'cas_masquerade_eduPersonPrimaryAffiliation' => env('CAS_MASQUERADE_eduPersonPrimaryAffiliation'),
+    'cas_masquerade_givenName' => env('CAS_MASQUERADE_givenName'),
+    'cas_masquerade_sn' => env('CAS_MASQUERADE_sn'),
+    'cas_masquerade_authnContextClass' => env('CAS_MASQUERADE_authnContextClass'),
+    'cas_masquerade_authenticationDate' => env('CAS_MASQUERADE_authenticationDate'),
+    'cas_masquerade_eduPersonScopedAffiliation' => [env('CAS_MASQUERADE_eduPersonScopedAffiliation')],
+    'cas_masquerade_gtAccessCardNumber' => env('CAS_MASQUERADE_gtAccessCardNumber'),
```
evan10s, line 188: can you add this env var in .env.example for completeness?

evan10s (commented):
Changing GTID to nullable seems like it's going to have some unexpected downstream effects. I can't really think of a concrete example at the moment though

app/Http/Controllers/AttendanceController.php
```diff
@@ -53,6 +54,16 @@ public function store(StoreAttendanceRequest $request): JsonResponse
         // Variables for comparison below
         $date = $request->input('created_at', date('Y-m-d'));
         $gtid = $request->input('gtid');
+
+        // if an access card number is provided without a GTID, try to find the matching GTID
+        if ($request->input('gtid') === null && $request->input('access_card_number') !== null) {
+            if (AccessCard::where('access_card_number', '=', $request->input('access_card_number'))->exists()) {
+                $card = AccessCard::where('access_card_number', '=', $request->input('access_card_number'))->sole();
```
JoshOldenburg, line 61: I think you could combine this all into one `find()` call.

JoshOldenburg (commented):
Looks good so far.

app/Http/Controllers/AttendanceController.php
kberzinch: Nope https://github.com/laravel/framework/blob/11.x/src%2FIlluminate%2FDatabase%2FQuery%2FBuilder.php#L2900

PR #4733, opened 2024-07-28
Record method for distributing merchandise

### LINKED ISSUE #4731: Add distribution source to distribution model and API

kberzinch:
Closes #4731


app/Http/Controllers/MerchandiseController.php
```diff
@@ -69,7 +69,7 @@ public function getDistribution(Merchandise $merchandise, User $user): JsonRespo
                     'status' => 'error',
                     'message' => self::NO_DTM,
                 ],
-                status: 400
+                status: 404
```
evan10s, line 72: I was getting a stack trace back on stage for a random non-existent GTID. From this it looks like that case should have been returning a JSON error though

evan10s (approved):

app/Http/Controllers/MerchandiseController.php
kberzinch: That would be coming from the [route model binding](https://laravel.com/docs/11.x/routing#customizing-the-resolution-logic) here: https://github.com/RoboJackets/apiary/blob/143d2979505b7d160cbf2d7b4dd153492fc2ef9e/app/Providers/AppServiceProvider.php#L146

The controller method is either invoked with a fully-hydrated `User` model or doesn't get invoked at all.

PR #4823, opened 2024-09-22
Added migration - add class standings as of 2024-09-22 - Issue 4822

jvogt23:
Added migration that populates class_standings. This change is to prevent exceptions when logging into Sandbox with CAS

### DIFF
--- a/database/migrations/2024_09_22_161700_add_class_standings.php
+++ b/database/migrations/2024_09_22_161700_add_class_standings.php
@@ -50,7 +50,7 @@ return new class extends Migration
 
     public function down(): void
     {
-        Schema::table('class_standings', function($table) {
+        Schema::table('class_standings', function ($table) {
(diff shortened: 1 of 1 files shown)


database/migrations/2024_09_22_161700_add_class_standings.php
```diff
@@ -0,0 +1,57 @@
...
+
+declare(strict_types=1);
+
+use App\Models\ClassStanding;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
```
kberzinch, line 11: For the sake of the IDs also being in order, please reorder the inserts by ranking order

kberzinch (changes requested):

kberzinch (approved):

PR #4894, opened 2024-11-01
Fix Stick-To-Team checkbox on badge-in terminal

### LINKED ISSUE #4780: Kiosk not sticking to same team
The "stick to team" checkbox does not seem to work (at least not for RoboCup). We have tried it multiple times and every time it just reverts back to the home screen and each person has to click on RoboCup again.

jvogt23:
Instead of firing a new SweetAlert event, increases the timer on the current event to allow multiple badge-ins under the same team. Resolves #4780 

P.S. Originally, the addition to the docker-compose.yml that adds Passport client secrets and IDs created a duplicate record error when rebuilding the docker container multiple times. The line I added resolves this issue, but I am not sure it is the best solution to that problem. If a reviewer believes it will cause problems, please let me know and I will pursue other solutions if available.


resources/js/components/attendance/AttendanceKiosk.vue
```diff
@@ -400,26 +400,22 @@
...
-                                  });
-                                }
-                            });
-                            if (!self.stickToTeam) {
+                                });
                                 this.clearFields();
-                            } else {
+                            } else { // Do not fire a swal event - When the badge-in swal closes, it clears the current team
                                 this.clearGTID();
+                                const teamToStickTo = this.teams.filter(team => team.id.toString() === this.attendance.attendable_id);
```
evan10s, line 415: Why do we need to reset teamToStickTo here?

JoshOldenburg (dismissed):

kberzinch (approved):

PR #4956, opened 2024-12-09
Unsaved Change Warning on UpdateUserProfile - #1060

jvogt23:
Uses vanilla JS and should be replaced with navigation guards if we switch to vue router. Solves #1060


resources/js/components/UserEditForm.vue
```diff
@@ -326,7 +335,18 @@ export default {
...
+  beforeDestroy() {
+    window.removeEventListener('beforeunload', this.onLeaveAttempt);
+  },
   methods: {
+    onFormChange() {
+      this.formChanged = true;
+    },
+    onLeaveAttempt(event) {
+      if (this.formChanged) {
+        event.returnValue = "Are you sure you want to leave? Any unsaved changes will be lost.";
```
evan10s, line 347: I thought the prompt choices, at least in chrome, are ok/cancel, which doesn't align with the question asked here

evan10s:
we should deploy to stage and test before merging, have you done that before?

jvogt23:
> we should deploy to apiary-test to test before merging, have you done that before?

I have not.

evan10s:
did we end up deploying this to test? I mostly just want to try this for myself and confirm the behavior, but can't remember if we did that already

jvogt23:
> did we end up deploying this to test? I mostly just want to try this for myself and confirm the behavior, but can't remember if we did that already

I'm fairly sure I deployed this to the test environment. if it's not there I can redo the deployment

evan10s:
> > did we end up deploying this to test? I mostly just want to try this for myself and confirm the behavior, but can't remember if we did that already
> 
> I'm fairly sure I deployed this to the test environment. if it's not there I can redo the deployment

yeah let's deploy and double-check but probably good to merge after that

evan10s (approved):

PR #4968, opened 2024-12-14
Clarify help text on trip assignment creation

kberzinch:
https://apiary-sandbox.robojackets.org/nova/resources/trip-assignments/new

### DIFF
--- a/resources/views/nova/help/travel/assignment/member.blade.php
+++ b/resources/views/nova/help/travel/assignment/member.blade.php
@@ -1,0 +1,1 @@
+Only members that have recently paid dues are shown here by default. If you need to create an assignment for a member that hasn't paid dues, go to their user details page and click <strong>Create Trip Assignment</strong> there.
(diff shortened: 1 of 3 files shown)


resources/views/nova/help/travel/assignment/member.blade.php
```diff
@@ -0,0 +1 @@
+Only members that have recently paid dues are shown here by default. If you need to create an assignment for a member that hasn't paid dues, go to their user details page and click <strong>Create Trip Assignment</strong> there.
```
evan10s, line 1: maybe remove "by default" for brevity?

also is "recently paid dues" simplifying some complicated logic? is it anyone considered active (vs. access active but not active? I just feel like the wording could be more precise

resources/views/nova/help/travel/assignment/trip.blade.php
```diff
@@ -1 +1 @@
-Only upcoming trips are shown here by default. If you need to create an assignment for a past trip, go to the trip details page and click <strong>Create Trip Assignment</strong> there.
+Only upcoming trips in draft status are shown here by default. If you need to create an assignment for a past trip, go to the trip details page and click <strong>Create Trip Assignment</strong> there.
```
evan10s, line 1: same here re "by default" unless I'm missing something

evan10s (approved):

resources/views/nova/help/travel/assignment/member.blade.php
kberzinch: It is just access active, yes, but I don't think that's helpful information for non-technical users.

PR #5102, opened 2025-03-29
Expose primary team for a user via API

kberzinch:


app/Models/User.php
```diff
@@ -1159,6 +1160,28 @@ static function (Builder $query): void {
...
+            ->leftJoin('teams', static function (JoinClause $join): void {
+                $join->on('teams.id', '=', 'attendance.attendable_id')
+                    ->where('attendable', '=', true);
+            })
+            ->where('attendable', '=', true)
+            ->groupBy('attendable_id')
+            ->select(
+                'attendable_id',
+                DB::raw('count(*) as count'),
+                DB::raw('\''.Team::getMorphClassStatic().'\' as attendable_type')
```
JoshOldenburg, line 1176: Feels super weird to have code directly in raw SQL but I don't have a better suggestion.

JoshOldenburg (approved):
Lgtm

app/Models/User.php
kberzinch: This is just to grab a string value

jvogt23 (commented):
LGTM

Issue #3939, opened 2023-09-17
Add action to manually create user from attendance record


kberzinch:
This comes up occasionally when the automatic BuzzAPI fetch fails

kberzinch:
Also i guess a lens to find these records in the first place

PR #5110, opened 2025-04-04
#3939 Add action to create user from attendance record.

jvogt23:
Created a Nova action to add a user based on an attendance record. Currently contains the minimum fields required to create a new user; can add more if requested. Fields are as follows
First/Last name
GTID
GT Email
Middle Name (optional)
Username
Reason for creation (can be made into autofilling field if needed)

To test, tap in (or manually type) attendance with a GTID not linked to a student, then attempt to run the action on the attendance record.


app/Nova/Actions/CreateUserFromAttendance.php
kberzinch, line 1: You don't need to prompt for all of these fields as input, they're available from BuzzAPI. Just invoke [CreateOrUpdateUserFromBuzzAPI](https://github.com/RoboJackets/apiary/blob/main/app/Jobs/CreateOrUpdateUserFromBuzzAPI.php).

kberzinch (changes requested):

app/Nova/Actions/CreateUserFromAttendance.php
jvogt23: So is this action just to re-invoke buzzapi if it fails the first time? Will we ever need a fully manual user creation?

kberzinch: Yes, it's just to retry the BuzzAPI call. If we need to manually create a user from scratch then there's already a screen for that in Nova.

jvogt23: Changed to a buzzapi call. Let me know if there's any problems with it. Probably needs to be put in the sandbox for testing if we aren't completely sure

kberzinch, line 1: Fix lint issues annotated on this file.

app/Nova/Attendance.php
kberzinch, line 1: Only show the action if there isn't already a user associated.

kberzinch (changes requested):

app/Nova/Actions/CreateUserFromAttendance.php
jvogt23: PHPStan linter seems to be broken; it's flagging a lot of files that I never touched.

kberzinch: Yeah, don't worry about those

app/Nova/Actions/CreateUserFromAttendance.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @return array<int, \Laravel\Nova\Fields\Field>
+     */
+    public function fields(NovaRequest $request): array
+    {
+        return [
+            Text::make('Reason for Creation')
+                ->maxlength(255)
+                ->enforceMaxLength()
+                ->default('Created From Attendance Record'),
```
kberzinch, line 47: I don't think this needs to be user input - you can just use the same reason as `AttendanceObserver`, or a different hardcoded reason if you'd like. https://github.com/RoboJackets/apiary/blob/2f51275227a613377efa0aeedf4d7ffe3b3d653b/app/Observers/AttendanceObserver.php#L19-L23
The creation reason isn't used for anything in particular, mostly just akin to an audit trail.

app/Nova/Actions/CreateUserFromAttendance.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        if ($models->count() > 1) {
+            return Action::danger('Cannot create a user from more than one attendance record.');
+        }
+        if (User::where('gtid', $models->sole()->gtid)->exists()) {
+            return Action::danger('User already exists for this attendance record.');
+        }
+        $gtid = $models->sole()->gtid;
+        try {
+            CreateOrUpdateUserFromBuzzAPI::dispatch('gtid', $gtid, $fields->reason_for_creation);
```
kberzinch, line 68: Since this is UI-driven behavior, it's better to make the BuzzAPI call synchronously, so the user can get immediate feedback.
```suggestion
            CreateOrUpdateUserFromBuzzAPI::dispatchSync('gtid', $gtid, $fields->reason_for_creation);
```

This is similar to what we do in `RefreshFromGTED`. https://github.com/RoboJackets/apiary/blob/2f51275227a613377efa0aeedf4d7ffe3b3d653b/app/Nova/Actions/RefreshFromGTED.php#L42-L46

kberzinch (changes requested):

kberzinch (approved):

PR #5130, opened 2025-04-22
Add help link to header

JoshOldenburg:


resources/views/layouts/app.blade.php
```diff
@@ -53,6 +53,12 @@
             </li>
             @endcan
 
+            @can('access-nova')
+            <li class="nav-item">
+              <a class="nav-link" href="/docs/">Help</a>
```
kberzinch, line 58: I would rather label this "Docs" - "Help" suggests a broader support function than just documentation IMHO.

kberzinch (approved):

Issue #5322, opened 2025-10-02
SponsorLoginController


jvogt23:
Create a SponsorController in pursuit of a Sponsor login page.

Sponsors require one-time password login. Workflow is as follows:
- Sponsor writes email
- Email is validated to ensure it is valid, its domain is in the list of SponsorDomains (see the Sponsor and SponsorDomain models and their associated issue on GitHub #5310), and it does not show signs of attempts to foil the system. Laravel and PHP have built-in validation rules to do this.
- OTP is sent to that email if valid, if not, error message is displayed on screen. If valid, screen updates with a password prompt.
- If password is correct, user logged in.

This will likely require throttling, expiring passwords, and login attempt logging.

PR #5341, opened 2025-10-12
Sponsor user - Supports #5322

jvogt23:
Added SponsorUser: minimal user model for sponsors that contains an email field so that spatie/laravel-one-time-passwords can send OTPs to sponsors.

Also added the laravel-one-time-passwords library.


LightningBoltz21 (dismissed):

kberzinch (dismissed):

config/one-time-passwords.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+return [
+    /*
+     * one-time passwords should be consumed within this number of minutes
+     */
+    'default_expires_in_minutes' => 2,
```
evan10s: Idk if we're overriding the default but 10 minutes is a more reasonable default imo/from what I've seen

evan10s (approved):

config/one-time-passwords.php
jvogt23: Default is 2 from this library but I can change it to 10

PR #5361, opened 2025-10-21
Check for team exists before calculate total members count

raphaelts3:
This PR aims to fix the issue https://github.com/RoboJackets/apiary/issues/3493 by checking if the team exists before trying to get total members count, and if it does not exists, return 0.

Closes https://github.com/RoboJackets/apiary/issues/3493 if approved.

### DIFF
--- a/app/Nova/Metrics/TotalTeamMembers.php
+++ b/app/Nova/Metrics/TotalTeamMembers.php
@@ -23,7 +23,11 @@ class TotalTeamMembers extends Value
      */
     public function calculate(Request $request): ValueResult
     {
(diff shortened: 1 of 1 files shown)


kberzinch (changes requested):
Thank you for your contribution! There is a simpler way to fix this since the issue is only triggered on deleted teams - you can use the [`withTrashed`](https://laravel.com/docs/12.x/eloquent#including-soft-deleted-models) method for Eloquent to return deleted models as part of the query.

raphaelts3:
@kberzinch thanks for the feedback! Fixed and rebased.

PR #5372, opened 2025-10-29
Sponsor Login Controller

LightningBoltz21:
Fixes issue #5322. NOTE: the functionality of the controller has not been tested yet, but I'm making a PR so what I have so far can be reviewed.

### DIFF
--- a/app/Nova/SponsorUser.php
+++ b/app/Nova/SponsorUser.php
@@ -79,6 +79,11 @@ class SponsorUser extends Resource
                 ->creationRules('unique:sponsor_users,email')
                 ->updateRules('unique:sponsor_users,email,{{resourceId}}')
                 ->sortable(),
+            Text::make('Email Suppression Reason', 'email_suppression_reason')
+                ->nullable()
+                ->readonly()
+                ->hideWhenCreating()
+                ->hideFromIndex(),
         ];
     }
 }


kberzinch (changes requested):

app/Http/Controllers/SponsorLoginController.php
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        // Verify OTP using Spatie
+        $result = $sponsorUser->attemptLoginUsingOneTimePassword((string) $request->input('otp'));
+        if (! $result->isOk()) {
+            return $this->errorResponse('Invalid OTP', $result->validationMessage());
+        }
+
+        // Save new user to database after successful OTP verification and sponsor check
+        if (! $sponsorUser->exists) {
```
jvogt23: The if (! sponsorUser->exists) statement is no longer needed because the sponsor user must be saved at an earlier step

app/Http/Controllers/SponsorLoginController.php
```diff
@@ -0,0 +1,144 @@
...
+            $sponsorUser->save();
+        }
+
+        // Establish authenticated session using Laravel's Auth facade
+        $request->session()->regenerate();
```
jvogt23: Lines 109, 113, and 115 I believe can be replaced by Auth::login(), and they don't need to be done alongside it.

jvogt23 (changes requested):

app/Http/Controllers/SponsorLoginController.php
```diff
@@ -0,0 +1,142 @@
...
+        $result = $sponsorUser->attemptLoginUsingOneTimePassword($otp);
+        if (! $result->isOk()) {
+            return $this->errorResponse('Invalid OTP', $result->validationMessage());
+        }
+
+        // Save user after successful OTP verification
+        $sponsorUser->save();
+
+        // Establish authenticated session using Laravel's Auth facade
+        // $request->session()->regenerate();
```
kberzinch: `regenerate()` is already called inside `Auth::login()` so we don't need to call it again here: https://github.com/laravel/framework/blob/a6d9e4bc9bb09dca0ccc6efdf348c6aca5f71766/src/Illuminate/Auth/SessionGuard.php#L578

Remove this commented line

kberzinch (changes requested):
Overall looks good, couple cleanup items

app/Http/Controllers/SponsorLoginController.php
```diff
@@ -0,0 +1,140 @@
...
+        if (! $sponsorDomain) {
+            return false;
+        }
+
+        return $sponsorDomain->sponsor && $sponsorDomain->sponsor->active();
+    }
+
+    private function errorResponse(string $title, string $message, int $status = 422): JsonResponse
+    {
+        // Dummy comment to force a build on Github site.
```
kberzinch: Please remove unneeded comments.

kberzinch (changes requested):
Looks ok to me, one last note about comments

kberzinch (approved):

app/Http/Controllers/SponsorLoginController.php
```diff
@@ -0,0 +1,139 @@
...
+                'string',
+                'email:rfc,strict,dns,spoof',
+                'max:255',
+            ],
+        ]);
+
+        // Read value - cast to string since validation guarantees it
+        $email = (string) $request->input('email');
+
+        // Check if domain is valid and sponsor is active; if not, return JSON error
```
evan10s: I would gently suggest removing comments that directly restate the logic in the code; this is above the amount of comments generally found in this codebase

evan10s (commented):
Would be good to test this on stage before merging

PR #5588, opened 2026-04-17
Kiosk Card Format Updates

### LINKED ISSUE #2749: Add support for Mobile Credential for kiosk

RossIV:
Fix #2749 - Update kiosk card data formats to all known current permutations including EV1 + EV3 Plastic and Mobile from various different readers/formats

This is backwards-compatible with the current reader format on the kiosk in the shop (which presents just GTID), so merging/deploying to production should not cause any issues.

The kiosk will now accept the following formats:
- GTID-only: `902900001`
- Mobile credential: sixteen-digit number starting with 601770 (e.g. `6017700010000123`)
- Plastic credential: 6-digit, 7-digit or 9-digit zero-padded (e.g. `800001`, `1000003`, `000800001`, `001000003`)
- Transact MRD5 custom format for RoboJackets: 
    `GTID|CardNumber|CardType|`
    Mobile Credential: `|6017700010000123|MobileA|`
    Plastic Credential: `902900001|800001|DESFire|`
- Transact Reader (MRD5, PS4101, maybe TWN4?) Raw (Non-Customized) Plastic Card Format
    `1570=GTID=00=RawCardNumber` (e.g. `1570=902900001=00=6017700008000010`)

Other changes:
- The "admin mode" trigger of presenting a credential belonging to a user with admin rights without selecting a team **only works with a GTID presentation** since that's what the search endpoint on the backend requires. Perhaps a later update could allow for searching by card number there in addition.

### DIFF
--- a/resources/js/components/attendance/AttendanceKiosk.vue
+++ b/resources/js/components/attendance/AttendanceKiosk.vue
@@ -296,8 +296,25 @@
             },
             submit() {
                 // Check for lack of team selection
-                if (this.attendance.attendable_id === '' && this.attendance.gtid !== '') {
-                    // We have a valid card read with GTID and no team picked, check if user is an admin for hidden menu access
+                if (this.attendance.attendable_id === '') {
+                    // We have a valid card read with and no team picked, check if user is an admin for hidden menu access
+                    if (this.attendance.gtid !== '') {
+                        // If there's no GTID, we can't query admin status, so assume they're not.
+                        this.hasError = true;
+                        this.feedback = '';
+                        this.clearFields();
+                        new Audio(this.sounds.error).play()
+                        Swal.fire({
+                          title: 'Whoops!',
+                          text: 'Select a team before tapping your BuzzCard',
+                          icon: 'warning',
+                          timer: 2500,
+                          timerProgressBar: true,
+                          showConfirmButton: false
+                        });
+                        return false;
+                    }
+
                     axios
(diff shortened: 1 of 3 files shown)


resources/js/components/attendance/AttendanceKiosk.vue
```diff
@@ -287,8 +296,8 @@
             },
             submit() {
                 // Check for lack of team selection
-                if (this.attendance.attendable_id === '') {
-                    // We have a valid card read and no team picked, check if user is an admin for hidden menu access
+                if (this.attendance.attendable_id === '' && this.attendance.gtid !== '') {
+                    // We have a valid card read with GTID and no team picked, check if user is an admin for hidden menu access
                     axios
                         .get(this.usersBaseUrl + "/search", {
```
kberzinch: While we're poking around here - this doesn't actually need to use the search endpoint, you can use the GTID in place of the user ID for the normal GET endpoint

kberzinch (commented):
Looks fine from a glance - please rebase to pick up a CI fix

resources/js/components/attendance/AttendanceKiosk.vue
kberzinch: The response format is different for user get vs search, you only get a single user back instead of a list

kberzinch (changes requested):

kberzinch (approved):

PR #5734, opened 2026-07-26
Add Record Attendance action in Nova for Teams and Events

### LINKED ISSUE #2542: Attendance widget does not work on team/event pages in Nova 4.x

RossIV:
## What

Adds a **Record Attendance** user action to the **Team** and **Event** detail views in Nova. It opens an interactive modal where an admin types or swipes a GTID or access card number; each entry is validated client-side, sent to the server for real validation/processing, and the result — including the matched user's name (or "Non-Member") — is shown inline. The modal stays open so multiple people can be checked in one after another, with a display-only per-session counter.

This gives us in-Nova attendance collection alongside the existing Remote Attendance Links, Kiosk, and mobile-app paths.

## How

- **Reuses the existing `POST /api/v1/attendance` endpoint** — no server-side attendance logic was changed. It already resolves an access card to a GTID, dedupes per day, kicks off async user creation via `AttendanceObserver` + BuzzAPI, and returns the matched attendee's name. The modal `POST`s to it directly via `Nova.request()`. A logged-in Nova admin authenticates through Passport's `laravel_token` cookie (`CreateFreshApiToken`, already in the web middleware) plus the `X-CSRF-TOKEN` header Nova's axios sends. Gated by the existing `create-attendance` permission.
- **`app/Nova/Actions/RecordAttendance.php`** — an `onlyOnDetail` action with `public $component = 'record-attendance'`, wired into `Team::actions()` and `Event::actions()` with `canSee`/`canRun` on `create-attendance`. On Team it stays inside the existing `attendable`-aware block.
- **`nova-components/RecordAttendanceModal/`** — a new **Vue 3** Nova asset package for the modal. Nova 5 runs Vue 3, while the main app bundle is Vue 2, so a component compiled through the app pipeline throws `_c` errors at runtime; it must be built as a separate `.vue({ version: 3 })` package like the existing modals. Registered from `NovaServiceProvider` via `Nova::script` (rather than a Composer path package) to avoid churning `composer.lock`. Styled to match native Nova action modals.
- **Shared credential parser** — extracted the kiosk's card-format detection into `resources/js/attendance/parseCredential.js`, now used by both the kiosk and the new modal so the two stay in sync. Fixed a latent bug where the Transact-raw plastic-card branch referenced an undefined `bbRawRegex` (should have been `tsRawRegex`).
- **Removed** the dead `resources/js/components/nova/CollectAttendance.vue` (the no-longer-used predecessor; zero references).
- **Dockerfile** — builds the new package in the `nova-components` stage (copying the shared parser in so its import resolves) and copies its `dist/` into the app image.

## Testing

Verified end-to-end in the local Docker environment:
- Action appears on Team/Event detail; modal matches native Nova styling (header, footer bar, primary/ghost buttons).
- New GTID → "Recorded — {name}" / "Non-Member", counter increments; same-day repeat → "Already recorded today — {name}", counter unchanged; each attendance row lands with `source = 'nova'` and the correct `recorded_by`/`attendable`.
- Only the most recent result is shown; the input auto-refocuses for the next entry.
- Kiosk regression: the extracted parser still handles every format (GTID, mobile, plastic incl. zero-padded, MRD5, Transact-raw — including the previously-broken branch).

## Reviewer notes

- `Event::actions()` already returns `[]` once `end_time` has passed, so the action is hidden on past events (consistent with the existing Create Remote Attendance Link action). Let me know if attendance should be recordable for just-ended events.
- The input carries `data-1p-ignore` / `data-lpignore` / `data-form-type="other"` so password managers don't offer credit-card autofill.
- PHP static analysis (Pint/PHPStan/Psalm) was not run locally — please run in CI/your environment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)
image
image
image

Closes #2542


app/Providers/NovaServiceProvider.php
```diff
@@ -32,6 +32,14 @@ public function boot(): void
 
         Nova::serving(static function (ServingNova $event): void {
             Nova::script('sentry', asset('js/nova.js'));
+
+            // The record-attendance modal is a Vue 3 component (Nova runs Vue 3, unlike the Vue 2 app
+            // bundle), so it is built separately in nova-components/RecordAttendanceModal. Registered
+            // here rather than via a Composer package to avoid a path-package for this app-owned action.
```
kberzinch: This is inconsistent with the other two custom components but I guess it's fine

kberzinch (dismissed):

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
```diff
@@ -0,0 +1,220 @@
...
+                    dusk="attendance-counter"
+                    title="Records created this session"
+                >
+                    {{ count }}
+                </span>
+            </h3>
+
+            <div class="px-8">
+                <p class="text-sm text-gray-500 dark:text-gray-400 mb-4">
+                    Swipe a BuzzCard or type a GTID or BuzzCard number, then press Enter.
```
evan10s: can we just say Swipe/Enter GTID? is the detail that a BuzzCard number (is that the access card number?) can also be used that important that it needs to be mentioned in all the places?

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
```diff
@@ -0,0 +1,220 @@
...
+                    autocapitalize="off"
+                    autocorrect="off"
+                    spellcheck="false"
+                    data-1p-ignore="true"
+                    data-lpignore="true"
+                    data-bwignore="true"
+                    data-form-type="other"
+                    :disabled="submitting"
+                    class="w-full form-control form-input form-input-bordered"
+                    placeholder="GTID or BuzzCard number"
```
evan10s: same here

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
```diff
@@ -0,0 +1,220 @@
...
+                .then(response => {
+                    const attendance = response.data.attendance;
+                    const name = attendance.attendee != null ? attendance.attendee.name : 'Non-Member';
+
+                    if (response.status === 201) {
+                        this.count += 1;
+                        this.showResult('success', 'Recorded — ' + name);
+                    } else {
+                        // 200: attendance already existed for today.
+                        this.showResult('duplicate', 'Already recorded today — ' + name);
```
evan10s: nit: In the app we don't distinguish this case and just treat it as successful. I don't think the distinction matters

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
```diff
@@ -0,0 +1,220 @@
...
+                    if (status === 403) {
+                        this.showResult('error', "You don't have permission to record attendance.");
+                    } else if (status === 422) {
+                        this.showResult('error', 'That does not look like a valid GTID or BuzzCard number.');
+                    } else {
+                        this.showResult(
+                            'error',
+                            'Unable to record attendance. Check your connection and try again.'
+                        );
+                    }
```
evan10s, line 201: We may want to report exceptions to sentry in this unhandled case, currently you are swallowing it

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
```diff
@@ -0,0 +1,220 @@
...
+
+<script>
+// The credential parser is shared with the attendance kiosk (resources/js/components/attendance/
+// AttendanceKiosk.vue) so both entry points detect card formats identically. It is bundled from the
+// app's resources during the build; the Dockerfile copies it into the nova-components build stage.
+import parseCredential from '../../../../../resources/js/attendance/parseCredential';
+
+export default {
+    props: {
+        working: Boolean,
```
evan10s, line 98: this doesn't seem to be used anywhere else

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
```diff
@@ -0,0 +1,220 @@
...
+        resourceName: { type: String, required: true },
+        action: { type: Object, required: true },
+        selectedResources: { type: [Array, String], required: true },
+        errors: { type: Object, required: true },
```
evan10s, line 102: this doesn't seem to be used anywhere else

resources/js/attendance/parseCredential.js
```diff
@@ -0,0 +1,78 @@
...
+export default function parseCredential(cardData) {
+    if (cardData === null || cardData === undefined) {
+        return null;
+    }
+
+    const value = String(cardData).trim();
+
+    if (value === '') {
+        return null;
+    }
```
evan10s, line 30: consider if these `return null`s should actually be raising exceptions or making sentry logs

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
kberzinch: Access card number support was dropped from the backend in https://github.com/RoboJackets/apiary/compare/9366c844c39b5fe5596f93650cecd7cdc6bfe2de...57a6b916681b65a8657eb80cff966c8ec54e178d

resources/js/attendance/parseCredential.js
RossIV: These could happen if whatever upstream didn't send a credential string to get parsed, which could be as simple as a false Enter key press without validation upstream. In that case, I don't think it'd be worth the Sentry noise. I'm open to other thoughts of how these could potentially get hit in such a way that it would be worth emphasizing.

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
RossIV: I don't understand the specific concern here. My understanding is this is standard Nova plumbing for modals. See also: Nova's `ConfirmActionModal` (not in this repo).

nova-components/RecordAttendanceModal/resources/js/components/RecordAttendance.vue
RossIV: See above

app/Nova/Actions/RecordAttendance.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace App\Nova\Actions;
+
+use Illuminate\Support\Collection;
+use Laravel\Nova\Actions\Action;
+use Laravel\Nova\Fields\ActionFields;
+
+/**
+ * Opens a custom modal (the `record-attendance` Vue component) on the Team/Event detail view for
+ * recording attendance by GTID or access card number. The modal talks directly to the attendance
```
kberzinch: Stray reference left to access card number here

app/Nova/Actions/RecordAttendance.php
```diff
@@ -0,0 +1,45 @@
...
+     * Perform the action on the given models.
+     *
+     * The record-attendance modal submits directly to the attendance API, so this is only reached if
+     * the action is somehow run through Nova's normal action pipeline.
+     *
+     * @param  \Illuminate\Support\Collection<int,\App\Models\Team|\App\Models\Event>  $models
+     */
+    public function handle(ActionFields $fields, Collection $models)
+    {
+        return Action::danger('Use the Record Attendance window to enter a GTID or access card number.');
```
kberzinch: And here

resources/js/components/attendance/AttendanceKiosk.vue
```diff
@@ -288,7 +238,21 @@
...
                     })
+                    return;
                 }
+
+                if (parsed.gtid !== null) {
+                    this.attendance.gtid = parsed.gtid;
+                }
+                if (parsed.access_card_number !== null) {
+                    this.attendance.access_card_number = parsed.access_card_number;
+                }
```
kberzinch: This needs to be refactored to throw an error if GTID is not available, and you can drop the `access_card_number` field at this point in the flow entirely (keep it in `parseCredential` if you want i guess?)

kberzinch (changes requested):
Couple spots still refer to access card numbers

kberzinch (dismissed):

resources/js/attendance/parseCredential.js
```diff
@@ -11,7 +11,7 @@
 // in sync. The Nova modal is built in an isolated stage; the Dockerfile copies this file in so its
 // relative import resolves.
 
-const mrd5Regex = /\|?(\d*)\|(\d+)\|(\w+)\|/;
+const tsCustomRegex = /\|?(\d*)\|(\d*)\|(\w+)\|?/;
```
evan10s, line 14: Why the optional | at the start and end?

evan10s (dismissed):

kberzinch (approved):

resources/js/attendance/parseCredential.js
RossIV: At one time I had a PS4101 configured to have `|` at the beginning and end, though that's no longer the case. The PS4101 at the shop today just does GTID, and the MRD5s don't have the `|` at either end, so this could probably be updated.

PR #5757, opened 2026-08-06
Create device inventory system

kberzinch:
Add a model and controller to collect basic information about devices, including hardware, firmware, and software versions, battery percentage, and the most recent user.

### DIFF
--- a/app/Http/Requests/StoreDeviceRequest.php
+++ b/app/Http/Requests/StoreDeviceRequest.php
@@ -10,6 +10,8 @@ class StoreDeviceRequest extends FormRequest
 {
     /**
      * Determine if the user is authorized to make this request.
+     *
+     * @psalm-pure
      */
     public function authorize(): bool
     {
@@ -20,6 +22,8 @@ class StoreDeviceRequest extends FormRequest
      * Get the validation rules that apply to the request.
      *
      * @return array<string, array<string>>
+     *
+     * @psalm-pure
      */
     public function rules(): array
     {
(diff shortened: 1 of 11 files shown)


app/Http/Requests/StoreDeviceRequest.php
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @psalm-pure
+     */
+    public function rules(): array
+    {
+        return [
+            'serial_number' => [
+                'required',
+                'digits:7',
+            ],
```
evan10s, line 34: is this only intended for MRD5s or would we also want this to extend to the loaner Android devices

evan10s (approved):

app/Http/Requests/StoreDeviceRequest.php
kberzinch: I was thinking about that. The Android devices already have MDM set up so I think this would be redundant, but I'd be open to extending it if you think it would be helpful. The MRD5S also expose manufacturer and model number strings through the standard BLE services, so we could make a composite primary key or something like that.

kberzinch: The Androids are also _supposed_ to be signed into the `apiary-mobile` account, so the user tagging may be less helpful

kberzinch: Reading e.g. IMEI or serial number on Android is not trivial either, we would likely need to have special handling for when the app is running in kiosk mode under MDM

Issue #5456, opened 2026-01-24
Resume model


jvogt23:
A Resume model is needed. Every user can have up to one Resume active at any time.

The resume model should contain a file path to the storage location of the user's actual resume.
Any filterable attributes not already present in the User model should also be included. The basic filterable attributes right now should be year in school (or graduation year), major, class standing. These attributes are in the User model already and can be referenced that way.

The model must also contain a toSearchableArray() which extracts the text from the resume. See https://github.com/RoboJackets/loop/blob/main/app/Models/Attachment.php#L131-L188
This method is used when indexing the model for searching.

PR #5768, opened 2026-08-11
Resume model

jvogt23:
Create a Resume model with the attributes
- user_id
- extracted_text
- last_uploaded
- Default timestamps
and a toSearchableArray() that indexes the extracted text.

Note that functionality for creating an instance on resume upload has not been implemented, nor has text extraction, at this time.

This PR covers issue #5456.


database/migrations/2026_04_18_160855_create_resumes_table.php
```diff
@@ -0,0 +1,38 @@
...
+            $table->unsignedInteger('user_id');
+            $table->text('extracted_text')->nullable();
+            $table->timestamp('last_uploaded_at')->nullable();
+            $table->timestamps();
+
+            $table->unique('user_id');
+            $table->foreign('user_id')
+                ->references('id')
+                ->on('users')
+                ->cascadeOnDelete();
```
kberzinch: You can use [`foreignIdFor`](https://laravel.com/docs/13.x/migrations#column-method-foreignIdFor) here.

database/migrations/2026_04_18_160855_create_resumes_table.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('resumes', static function (Blueprint $table) {
+            $table->id();
+            $table->unsignedInteger('user_id');
+            $table->text('extracted_text')->nullable();
```
kberzinch: Why do you want to store the extracted text in the database, instead of only in Meilisearch?

kberzinch (changes requested):

database/migrations/2026_04_18_160855_create_resumes_table.php
jvogt23: From what I understand, the way you store items in Meilisearch is through Laravel Scout and a toSearchableArray(). How would I store the extracted text using toSearchableArray without first having it in the db? Can you point me to an example model that stores an attribute or attributes only in meilisearch?

Edit: Would we instead want to have toSearchableArray run text extraction at index time?

kberzinch: Many attributes on the user model exist only in meilisearch and not in the database, they are separate schemas

Or look at models in the loop repo for a closer comparison to what you're doing here

jvogt23: After looking at the User model, I see what you mean. I've removed it from the schema and added a todo comment showing that the text should be extracted at index time. I'll probably need either Apache Tika or a PHP library called PdfParser to do this, whichever is more convenient for everyone involved. (Is Tika loaded into Apiary right now?)

kberzinch: Tika is not currently available but should be easy to copy-paste from loop

database/migrations/2026_04_18_160855_create_resumes_table.php
```diff
@@ -0,0 +1,33 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('resumes', static function (Blueprint $table) {
+            $table->id();
+            $table->timestamp('last_uploaded_at')->nullable();
```
kberzinch: Is there ever going to be a difference between `last_uploaded_at` and `updated_at`?

jvogt23: In fairness, it may not, as any update to the model would automatically update updated_at. However I left it in so the name of the column would make sense. I can remove it or make it an alias of updated_at if it is redundant.

kberzinch: I would just drop this column and use updated_at, this is the resume table so I think a plain update timestamp should be fine

kberzinch (approved):

Issue #5677, opened 2026-06-14
Add text extraction from Resume PDFs


jvogt23:
Upon upload, student resumes should have their text extracted using one of the following tools:

- smalot: A PDF parser library built into PHP that can be installed directly into the app
- Apache Tika: An external program that can be used as a sidecar to the application

Smalot PDFParser is more lightweight and will be easier to install, while Apache Tika is more feature-rich.

After a library is selected, Resume text should be extracted during the upload process and stored in a manner that facilitates vector and/or keyword search.

PR #5773, opened 2026-08-17
Feature/text extraction

jvogt23:
Registered Apache Tika as a service in the Nomad configuration. Added code for text extraction during search indexing in the Resume model.

This change is for issue #5677.


app/Models/Resume.php
```diff
@@ -116,22 +120,66 @@ public function getFilePathAttribute(): string
         return Storage::disk(self::storageDisk())->path($this->storagePath());
     }
 
+    /**
+     * Modify the collection of models being made searchable.
+     */
+    public function makeSearchableUsing(Collection $models): Collection
+    {
+        return $models->load('user');
+    }
```
kberzinch: I don't think you need this here?

app/Models/Resume.php
```diff
@@ -116,22 +120,66 @@ public function getFilePathAttribute(): string
...
      * @return array<string,int|string>
      */
     public function toSearchableArray(): array
     {
+        $file_path = $this->file_path;
+
+        $full_text = '';
+
+        if (Storage::disk('local')->exists($file_path) && Storage::disk('local')->size($file_path) > 0) {
+            $file_hash = hash_file('sha512', Storage::disk('local')->path($file_path));
```
kberzinch: You probably don't need all of this caching logic, Loop has many duplicate documents for complicated reasons but I don't think that applies here.

kberzinch (changes requested):

kberzinch (approved):

PR #5779, opened 2026-08-21
Migrate resumes

### LINKED ISSUE #5457: Migrate database for Resume
Migrating the database for the Resume model may be more intensive than a standard database migration. In addition to creating the table, we will need to populate the table based on the files already present in the resume folder.

jvogt23:
Resolves #5457. Added a migration to upsert resumes and a command to create a new Resume model instance when users upload resumes.


app/Http/Controllers/ResumeController.php
```diff
@@ -216,6 +217,14 @@ public function store(string $id, StoreResumeRequest $request)
             $user->resume_date = now();
             $user->save();
 
+            if (! Resume::where('user_id', $user->id)->exists()) {
+                Resume::create([
+                    'user_id' => $user->id,
+                ]);
+            } else {
+                Resume::where('user_id', $user->id)->first()->touch();
+            }
```
kberzinch: Use [`firstOrCreate`](https://laravel.com/framework/docs/13.x/eloquent#retrieving-or-creating-models) here

kberzinch (changes requested):

app/Nova/Resume.php
```diff
@@ -0,0 +1,81 @@
...
+     * Get the fields displayed by the resource.
+     */
+    #[\Override]
+    public function fields(NovaRequest $request): array
+    {
+        return [
+            BelongsTo::make('User', 'user', User::class)
+                ->rules('required')
+                ->sortable(),
+            Text::make('Last Uploaded', 'updated_at')
```
kberzinch:
```suggestion
            DateTime::make('Last Uploaded', 'updated_at')
```

app/Nova/Resume.php
kberzinch, line 1: You also likely should create a corresponding policy for this resource: https://nova.laravel.com/docs/v5/resources/authorization#policies

kberzinch (changes requested):

app/Nova/Resume.php
jvogt23: Who should be allowed to view Resumes? and should CRUD operations be allowed at all, since the Resume model is mostly just there to make resumes searchable once they are uploaded?

kberzinch: There is an existing `read-users-resume` permission you should check. They should probably be readonly.

app/Http/Controllers/ResumeController.php
```diff
@@ -216,6 +217,9 @@ public function store(string $id, StoreResumeRequest $request)
             $user->resume_date = now();
             $user->save();
```
kberzinch: Are you planning to drop this column at some point?

app/Nova/Resume.php
kberzinch, line 1: Link this to the user resource using [`HasOne`](https://nova.laravel.com/docs/v5/resources/relationships#hasone)

kberzinch (changes requested):

app/Http/Controllers/ResumeController.php
jvogt23: I intend to do this in my next change, which will be updating the resume extraction method and the sponsor pages to use the Resume model. If it's better to do that here than in a separate PR I can go ahead with that.

kberzinch: IMO it should be part of this PR, as it stands, there are two separate copies of the resume state on this branch.

app/Models/Resume.php
```diff
@@ -113,9 +128,40 @@ public function getFileNameAttribute(): string
...
+
+    /**
+     * Get the is_active flag for the Resume.
+     *
+     * @psalm-mutation-free
+     */
+    public function getIsActiveAttribute(): bool
+    {
+        return self::where('id', $this->id)->active()->count() !== 0;
+    }
```
kberzinch: Why do you need an `is_active` attribute on the resume? Laravel generally does a good job with relationships, so I don't see why this would be necessary.

app/Models/Resume.php
```diff
@@ -126,39 +172,39 @@ public function getFilePathAttribute(): string
...
+                    'synchronous' => true,
+                ]
+            )->put(
+                '/tika',
+                [
+                    'body' => Storage::disk(self::storageDisk())->get($this->storage_path),
+                ]
+            )->getBody()->getContents();
+            if ($full_text === '') {
+                $full_text = 'Tika extraction did not work';
```
kberzinch: I don't know what you should do in this scenario but this is not it

app/Nova/Actions/ExportFilteredResumes.php
kberzinch, line 1: This action should maybe be moved to the `Resume` resource instead of the `User` resource?

https://github.com/RoboJackets/apiary/blob/3e2cf6a1f60cd6599f56e3decb472b3bd0cd6c88/app/Nova/User.php#L727-L755

app/Nova/User.php
```diff
@@ -434,7 +454,7 @@ public function fields(NovaRequest $request): array
                             return $request->user()->can('read-users-resume');
                         }),
 
-                    DateTime::make('Resume Uploaded At', 'resume_date')
+                    DateTime::make('Resume Uploaded At', $this->resume?->updated_at)
                         ->onlyOnDetail()
                         ->hideFromDetail(static fn (NovaRequest $r, AppModelsUser $u): bool => $u->is_service_account),
```
kberzinch: These fields should both move to the `Resume` resource

app/Policies/ResumePolicy.php
```diff
@@ -0,0 +1,93 @@
...
+class ResumePolicy
+{
+    use HandlesAuthorization;
+
+    /**
+     * Determine whether the user can view any models.
+     */
+    public function viewAny(User $user): bool
+    {
+        return $user->can('read-users-resume');
```
kberzinch, line 24: Check if they have any resume uploaded (so they can see their own)

app/Policies/ResumePolicy.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Determine whether the user can view the model.
+     */
+    public function view(User $user, Resume $resume): bool
+    {
+        return $user->can('read-users-resume');
```
kberzinch, line 36: Check if they are trying to view their own resume

config/scout.php
```diff
@@ -264,6 +264,45 @@
...
+                    'updated_at',
+                    'extracted_text',
+                ],
+                'searchableAttributes' => [
+                    'id',
+                    'user_id',
+                    'file_name',
+                    'file_path',
+                    'created_at',
+                    'updated_at',
```
kberzinch: Are you sure you want all of these fields to be searchable? https://www.meilisearch.com/docs/capabilities/full_text_search/how_to/configure_searchable_attributes

config/scout.php
```diff
@@ -264,6 +264,45 @@
...
+                    'extracted_text',
+                ],
+                'filterableAttributes' => [
+                    'id',
+                    'user_id',
+                    'file_name',
+                    'file_path',
+                    'created_at',
+                    'updated_at',
```
kberzinch: I do not see any usage of filtering on these attributes, please remove them from the index settings to improve index performance.

database/migrations/2026_08_20_204527_upsert_resumes.php
kberzinch, line 1: Add a separate migration to drop the `resume_date` column from the user table, because it is redundant after this migration is applied.

kberzinch (changes requested):

app/Models/Resume.php
jvogt23: This was part of earlier testing, and $full_text was initially intended to be left blank. I am also unsure what to do about this, not sure if it would be a good idea to just leave it blank either since it would then be failing silently.

kberzinch: Probably throw an error and we can look at the document manually

app/Models/Resume.php
jvogt23: Are you also asking whether we should keep scopeActive() on the Resume model? If so the logic for that would just be the same as before, checking whether the user is active, has paid dues, and is not a service account

kberzinch: Ah, I misread. "Active" is kind of a weird name for this property, we already use that for membership and access, and I think this is overloading it a bit.

Maybe something like "visible" or "exportable"?

app/Nova/Resume.php
```diff
@@ -0,0 +1,159 @@
...
+                ->rules('required')
+                ->sortable(),
+            File::make('Resume', 'storage_path')
+                ->path('resumes')
+                ->disk('local')
+                ->deletable(false)
+                ->onlyOnDetail()
+                ->canSee(static function (Request $request): bool {
+                    $r = AppModelsResume::find($request->resourceId);
+                    if ($r && $r->user->id === $request->user()->id) {
```
kberzinch: User ID types are weird right now, use [`Model::is`](https://laravel.com/framework/docs/13.x/eloquent#comparing-models) to be safe

app/Nova/Resume.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Get the actions available for the resource.
+     *
+     * @return array<\Laravel\Nova\Actions\Action>
+     */
+    #[\Override]
+    public function actions(NovaRequest $request): array
+    {
+        $exportResumes = $request->user()->can('read-users-resume') ?
```
kberzinch: Return from here instead of putting it inside a local, splatting it into another list, and then returning

app/Policies/ResumePolicy.php
```diff
@@ -0,0 +1,101 @@
...
+
+        return $user->can('read-users-resume');
+    }
+
+    /**
+     * Determine whether the user can view the model.
+     */
+    public function view(User $user, Resume $resume): bool
+    {
+        if ($resume->user->id === $user->id) {
```
kberzinch: Same as above, use [`Model::is`](https://laravel.com/framework/docs/13.x/eloquent#comparing-models) here

kberzinch (changes requested):
Looks functionally OK, a few code style notes

app/Nova/Resume.php
```diff
@@ -0,0 +1,155 @@
...
+                $raw = AppModelsResume::search('')
+                    ->where('id', $resume->id)
+                    ->raw();
+
+                return $raw['hits'][0]['extracted_text'] ?? '';
+            })
+                ->onlyOnDetail()
+                ->canSee(static function (Request $request): bool {
+                    $r = AppModelsResume::find($request->resourceId);
+                    if ($r && $r->user->id === $request->user()->id) {
```
kberzinch: Missed a spot here

kberzinch (changes requested):

database/migrations/2026_08_20_204527_upsert_resumes.php
```diff
@@ -0,0 +1,36 @@
...
+    public function up(): void
+    {
+        $resumes = Storage::disk('local')->files('resumes');
+        $resumes = array_map(static fn ($f) => pathinfo($f, PATHINFO_FILENAME), $resumes);
+
+        $users = User::whereIn('uid', $resumes)->get();
+        Resume::upsert(
+            $users->map(static fn ($user) => ['user_id' => $user->id])->all(),
+            ['user_id']
+        );
```
kberzinch: This will reset the resume date for everyone, which may be undesirable since `updated_at` is used for filtering here: https://github.com/RoboJackets/apiary/blob/b068e88611eab9320546b7b65c2a644c6d52e525/app/Nova/Actions/ExportFilteredResumes.php#L100-L102

kberzinch (dismissed):
Looks ok, one functional change I wanted to call out.

Also note:
- the database doesn't actually have a foreign key between resumes and users, `foreignIdFor` is just syntactic sugar for the column type
- there should also be a unique index on user id, assuming we only want a single resume object per user

database/migrations/2026_08_20_204527_upsert_resumes.php
kberzinch: Also used to show a dashboard card here: https://github.com/RoboJackets/apiary/blob/migrate-resumes/app%2FHttp%2FControllers%2FDashboardController.php#L96

jvogt23: May need the migrations undone and redone to test whether what I have written resolves this problem

database/migrations/2026_08_20_204527_upsert_resumes.php
```diff
@@ -0,0 +1,63 @@
...
+
+        DB::table('resumes')->orderBy('id')->chunk(200, static function ($rows) {
+            foreach ($rows as $row) {
+                DB::table('users')
+                    ->where('id', $row->user_id)
+                    ->update(['resume_date' => $row->updated_at]);
+            }
+        });
+
+        Resume::truncate();
```
kberzinch: Missing an import for this

kberzinch (changes requested):

database/migrations/2026_08_20_204527_upsert_resumes.php
```diff
@@ -0,0 +1,64 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('resumes', static function (Blueprint $table) {
+            $table->foreign('user_id')
+                ->references('id')
+                ->on('user')
```
kberzinch:
```suggestion
                ->on('users')
```

kberzinch (changes requested):

database/migrations/2026_08_20_204527_upsert_resumes.php
```diff
@@ -0,0 +1,64 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('resumes', static function (Blueprint $table) {
+            $table->foreign('user_id')
+                ->references('id')
```
kberzinch, line 20: These columns are not the same type so MySQL doesn't allow creating a foreign key here

kberzinch (changes requested):

database/migrations/2026_08_20_204527_upsert_resumes.php
jvogt23: Changed to unsignedInt

kberzinch (approved):
I think you have the only resume in test, but I ran the migration on a copy of the prod DB and it looks fine.

